PR #13, opened 2026-04-15
feat: simplify code

stephantul:
Index stats is now a property: everything it needs can be computed from the chunks we have in memory. The `index` classmethod was unnecessary. Calculating the index_root was also unnecessary.


Pringled (approved):
😍

PR #43, opened 2026-04-26
Exclude uv.lock

barseghyanartur:
Reduces the downloadable package size by quite a lot.


Pringled:
Nice find, thanks @barseghyanartur!

PR #47, opened 2026-04-28
refactor: dedup SembleIndex builders and MCP tool preamble

sebastianbreguel:
`SembleIndex.from_path` and `from_git` shared a 5-line tail (load_model fallback, `create_index_from_path` call, constructor). Extracted into a private `_build` classmethod so each public entrypoint focuses on its own concern: `from_path` validates filesystem paths, `from_git` handles cloning.

The `search` and `find_related` MCP tools shared a 7-line preamble for repo resolution and index lookup. Extracted into `_get_index_or_error` so each tool body focuses on its own logic.

The first commit applies a repo-wide ruff D213 docstring auto-fix that the pre-commit hook produced, plus a pre-existing mypy fix in the benchmarks (`_AsymmetricWrapper.encode` now matches the `Encoder` protocol). It also adds `.serena/` to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -205,3 +205,4 @@ cython_debug/
 marimo/_static/
 marimo/_lsp/
 __marimo__/
+.serena/
--- a/benchmarks/baselines/coderankembed.py
+++ b/benchmarks/baselines/coderankembed.py
@@ -3,6 +3,7 @@ import json
 import sys
 import time
 from collections import defaultdict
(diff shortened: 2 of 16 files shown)


stephantul:
Hey, thanks for bringing the pre-commit issue to our attention. We fixed it in a PR of our own. We're not really interested in the other changes, these seem like minor issues that we'll solve in a future refactor.

PR #63, opened 2026-05-04
feat: Add automatic reindexing for local paths

### LINKED ISSUE #56: How do I reindex?

Pringled:
When semble is started against a local directory, it now watches for file changes and automatically rebuilds the index in the background using `watchfiles`. This does do a full rebuild each time, which is fine since semble is extremely fast.  This is safe on rapid file changes as well since `watchfiles` automatically batches multiple changes and waits before rebuilding. The default for this is 1600ms (we can update this but I think this is fine, e.g. if you run pre-commit you change a lot of files instantly and it will only rebuild after waiting 1600ms).


src/semble/mcp.py
```diff
@@ -119,14 +122,39 @@ def __init__(self, model: Encoder) -> None:
...
+        self._watcher_task: asyncio.Task[None] | None = None
+
+    def _cache_key(self, source: str, ref: str | None = None) -> str:
+        """Compute the canonical cache key for a source."""
+        is_git = _is_git_url(source)
+        return (f"{source}@{ref}" if ref else source) if is_git else str(Path(source).resolve())
+
+    def evict(self, source: str) -> None:
+        """Evict the cache entry for the given source, if it exists."""
+        self._tasks.pop(self._cache_key(source), None)
```
stephantul: is non-eviction a sign something is going wrong? If so, it may be useful to show a warning here.

src/semble/mcp.py
```diff
@@ -119,14 +122,39 @@ def __init__(self, model: Encoder) -> None:
...
+
+    async def _watch_loop(self, path: str) -> None:
+        """Watch the given path for changes and evict the cache entry on changes."""
+        try:
+            async for _ in watchfiles.awatch(path):
+                self.evict(path)
+                try:
+                    await self.get(path)
+                except Exception:
+                    pass
```
stephantul: Idem (also below). If this exception appears, is something going wrong? should we note it?

src/semble/mcp.py
```diff
@@ -138,10 +166,11 @@ async def get(self, source: str, ref: str | None = None) -> SembleIndex:
...
-            if task.done():
+            if task.done() and self._tasks.get(cache_key) is task:
                 self._tasks.pop(cache_key, None)
             raise
         except Exception:
-            # Build failed: evict so the next caller can retry.
-            self._tasks.pop(cache_key, None)
+            # Only evict if this task hasn't already been replaced by evict()+get().
+            if self._tasks.get(cache_key) is task:
+                self._tasks.pop(cache_key, None)
```
stephantul: maybe use `evict` here? to get a single canonical reference.

src/semble/mcp.py
```diff
@@ -138,10 +166,11 @@ async def get(self, source: str, ref: str | None = None) -> SembleIndex:
         try:
             return await asyncio.shield(task)
         except asyncio.CancelledError:  # pragma: no cover
-            if task.done():
+            if task.done() and self._tasks.get(cache_key) is task:
                 self._tasks.pop(cache_key, None)
```
stephantul: same here, replace with `_evict`

src/semble/mcp.py
```diff
@@ -119,14 +122,39 @@ def __init__(self, model: Encoder) -> None:
         """Initialise an empty cache with a shared embedding model."""
         self._model = model
         self._tasks: dict[str, asyncio.Task[SembleIndex]] = {}
+        self._watcher_task: asyncio.Task[None] | None = None
+
+    def _cache_key(self, source: str, ref: str | None = None) -> str:
```
stephantul: this should be a verb, e.g. `_compute_cache_key`. This looks like an operation that caches a key or gets something from the cache.

stephantul (approved):

src/semble/mcp.py
Pringled: Agree, done

src/semble/mcp.py
Pringled: No this is expected and normal, evict is just called before every rebuild, and on the first build nothing needs to be evicted yet

Pringled: Ah didn't read your other comment yet, that one does need a logging line

src/semble/mcp.py
Pringled: Yes I'll add a logger warning here, that's an unexpected "error" (won't break anything but good to warn)

src/semble/mcp.py
Pringled: Agree

src/semble/mcp.py
Pringled: Agree

PR #69, opened 2026-05-05
fix: Bound resource usage and make git transports in MCP server safer

### LINKED ISSUE #59: Security: MCP server exposes local filesystem and lacks clone safety limits
The MCP server has two related issues that let an attacker (via prompt injection, a malicious document the agent
  reads, or a compromised tool output) coerce the agent into clones it shouldn't make.

  1. file:// URLs allow arbitrary local-repo read

  utils._GIT_URL_SCHEMES includes file://, so an LLM-supplied repo argument like file:///home/victim/some-private-repo
  is accepted by from_git and cloned via git clone. The repo's contents are then returned to the model through
  search/find_related. On developer machines, private git repos are usually trivial to find by path.

  ssh:// and SCP-form URLs (git@host:org/repo) also reach git clone unfiltered, enabling SSH connections to arbitrary
  internal hosts (useful for reconnaissance even though stdin is suppressed).

  Suggested fix: in mcp.py, restrict accepted schemes to https:// / http:// (or make non-https schemes opt-in via an env
   var like SEMBLE_ALLOW_LOCAL_GIT=1). The Python API can keep accepting all schemes — the boundary worth hardening is
  the MCP tool.

  2. No timeout, size, or cache limits → DoS

  SembleIndex.from_git calls subprocess.run(["git", "clone", ...]) with no timeout=. A slow or intentionally large
  remote blocks the MCP event loop indefinitely. There is also no max-file-size in chunker.chunk_file, no max-files cap
  in walk_files, and _IndexCache grows unbounded — chunks stay fully in RAM, with no LRU. A session that gets nudged to
  index several large repos can exhaust memory.

  Suggested fix:
  - subprocess.run(..., timeout=300) (configurable) and surface TimeoutExpired as RuntimeError.
  - --depth 1 --filter=blob:limit=10m on the clone, or a post-clone size check before indexing.
  - Cap _IndexCache to N entries with LRU eviction; expose a clear_cache MCP tool.
  - Skip files larger than e.g. 1 MB in chunk_file (configurable).

  3. (Bonus) MCP instructions encourage URL hallucination

  The server instructions string tells the model: "For questions about a library (e.g. a PyPI/npm package), resolve the
  GitHub URL from your training knowledge and pass it as repo." This actively encourages cloning typo-squatted or
  hallucinated repos and presenting their contents as authoritative. Recommend rewording to require an explicit URL from
   the user/context.

  Happy to send a PR if useful.

Pringled:
This PR addresses several issues from https://github.com/MinishLab/semble/issues/59:
- Reject unsafe git transport schemes (ssh://, file://, SCP-form) in MCP `repo` argument
- Cap `_IndexCache` at 10 entries (LRU) to prevent unbounded memory growth
- Skip files over 1 MB during indexing to prevent DoS via large files
- Add 60s timeout to `git clone` to prevent hung MCP server
- Fix MCP server instructions that encouraged the LLM to hallucinate GitHub URLs

The first concern (file:// URLs allow arbitrary local-repo read) is not addressed yet, as this changes the functionality of Semble (and arguably makes it worse). This is something for a followup PR.


src/semble/index/index.py
```diff
@@ -128,15 +128,17 @@ def from_git(
...
         :return: An indexed SembleIndex. Chunk file paths are repo-relative (e.g. ``src/foo.py``).
-        :raises RuntimeError: If git is not on PATH or the clone fails.
+        :raises RuntimeError: If git is not on PATH, the clone fails, or times out.
         """
         with tempfile.TemporaryDirectory() as tmp_dir:
             # `--` prevents `url` from being interpreted as a git option (e.g. `--upload-pack=...`).
             cmd = ["git", "clone", "--depth", "1", *(["--branch", ref] if ref else []), "--", url, tmp_dir]
             try:
-                result = subprocess.run(cmd, capture_output=True, text=True, stdin=subprocess.DEVNULL)
+                result = subprocess.run(cmd, capture_output=True, text=True, stdin=subprocess.DEVNULL, timeout=60)
```
stephantul: maybe consider making this a constant, that makes clear this is in seconds.

stephantul: Maybe also make this configurable using a flag? It could prevent users with very big repositories from using semble.

src/semble/mcp.py
```diff
@@ -124,7 +134,7 @@ class _IndexCache:
     def __init__(self, model: Encoder) -> None:
         """Initialise an empty cache with a shared embedding model."""
         self._model = model
-        self._tasks: dict[str, asyncio.Task[SembleIndex]] = {}
+        self._tasks: OrderedDict[str, asyncio.Task[SembleIndex]] = OrderedDict()
```
stephantul: This is an unrelated change right? I don't think the old version was actually different, or do you rely on ordering?

src/semble/mcp.py
```diff
@@ -17,21 +18,42 @@
...
+    default_source: str | None,
+    cache: _IndexCache,
+) -> SembleIndex:
+    """Return a cached index for a repo, rejecting unsafe git transport schemes."""
+    if repo is not None and _is_git_url(repo) and not repo.startswith(("https://", "http://")):
+        raise ValueError(f"Only https://, http://, or local directory paths are accepted as `repo`. Got: {repo!r}")
+    source = repo or default_source
+    if not source:
+        raise ValueError(
+            "No repo specified and no default index. Pass an https:// git URL or local directory path as `repo`."
```
stephantul: I'm assuming this error only mentions https because it's actually the preferred option? Same below, you only instruct the LLM to pass https, but http is also accepted.

stephantul (approved):

src/semble/index/index.py
Pringled: Yeah good point, will make it a constant. We could make it configurable via an env var imo. Via the CLI means we have to write it through the MCP server as well so I think we should just do an env var with this default.

src/semble/mcp.py
Pringled: Right, http is also fine, will update

src/semble/mcp.py
Pringled: Yeah for the LRU cache we need it, `move_to_end()` is only supported for OrderedDict afaik

PR #73, opened 2026-05-07
feat: Expose `include-text-files` via CLI

### LINKED ISSUE #72: [Feature]: add JSON & YAML to index
Currently, when I run `semble init`, it seems the JSON files are ignored.
I want to be able to index JSON & YAML files and search them.

Pringled:
This PR adds and `--include-text-files` to the CLI so it's in sync with the Python API. This resolves https://github.com/MinishLab/semble/issues/72. 

It's also added to the MCP args so a user can enable it by default if needed for any MCP server.


src/semble/cli.py
```diff
@@ -82,7 +97,12 @@ def _cli_main() -> None:
         _run_init(force=args.force)
         return
 
-    index = SembleIndex.from_git(args.path) if _is_git_url(args.path) else SembleIndex.from_path(args.path)
+    include_text = getattr(args, "include_text_files", False)
```
stephantul: I'm not familiar with this pattern being necessary. Is the arg added conditionally? For an action="store_true", the flag is set to false if it hasn't been added AFAIK.

stephantul (approved):
One small comment

src/semble/cli.py
Pringled: Oh you're right! Nice, will just change it to args.include_text_files then.

PR #74, opened 2026-05-07
feat: add ability to enable text file indexing via cli and mcp

sulaweyo:
I have worked on a bigger AI facing knowledge base and wanted to use semble to reduce my search times and token cost but found out that there is no way to enable the already present flag to index text files.
In this PR I have added the parameter handling to enable text file indexing via cli or mcp


Pringled:
@sulaweyo thanks, but there's already a PR for this which was just merged: https://github.com/MinishLab/semble/pull/73. Also this is not in line with our [contributing guidelines](https://github.com/MinishLab/semble/blob/main/CONTRIBUTING.md).

PR #76, opened 2026-05-07
feat: Add "semble savings" command

Pringled:
This PR adds `semble savings`, a CLI command that tracks and displays token savings across all searches.  Stats are recorded automatically to `~/.semble/savings.jsonl` on every search. There's also a verbose output that shows the calls per call type but it's not that interesting (yet) since we only have `search` and `find_related` atm.

Example output 💅 ✨:

```
  Semble Token Savings
  ════════════════════════════════════════════════════════════════
  Period        Calls   Savings
  ────────────────────────────────────────────────────────────────
  Today         42      [███████████████░]  ~58.4k tokens (95%)
  Last 7 days   287     [██████████████░░]  ~312.4k tokens (90%)
  All time      1.4k    [██████████████░░]  ~1.2M tokens (89%)
```


src/semble/index/index.py
```diff
@@ -104,6 +120,7 @@ def from_path(
         )
 
         index = SembleIndex(model, bm25, vicinity, chunks)
+        index._file_sizes = SembleIndex._compute_file_sizes(path, chunks)
```
stephantul: this can be `self._compute_file_sizes`. I would not call this here and just make it part of the `__init__`, or, if it is fast enough, just make it a property. There's also afaik no need to make this a staticmethod, doing

```
index.recompute_file_sizes()
```

is completely fine. I realize you do need the root, but the root could be part of the `SembleIndex`.

src/semble/index/index.py
```diff
@@ -171,7 +189,11 @@ def find_related(self, source: Chunk | SearchResult, *, top_k: int = 5) -> list[
         target = source.chunk if isinstance(source, SearchResult) else source
         selector = self._get_selector_vector(filter_languages=[target.language]) if target.language else None
         results = search_semantic(target.content, self.model, self._semantic_index, self.chunks, top_k + 1, selector)
-        return [r for r in results if r.chunk != target][:top_k]
+        results = [r for r in results if r.chunk != target][:top_k]
+        if self._file_sizes:
```
stephantul: I don't think this ever needs to be checked. It's probably just better to populate file_sizes here if it doesn't exist already.

src/semble/stats.py
```diff
@@ -0,0 +1,137 @@
...
+
+def save_search_stats(
+    results: list[SearchResult],
+    call_type: CallType,
+    file_sizes: dict[str, int],
+) -> None:
+    """Save stats about a search or find_related call to the stats file."""
+    try:
+        snippet_chars = sum(len(result.chunk.content) for result in results)
+        file_chars = sum(file_sizes.get(path, 0) for path in {result.chunk.file_path for result in results})
```
stephantul: It is probably better to disregard chunks for files for which we don't have size information. In fact, I think not having a specific file here points to some other failure.

src/semble/stats.py
```diff
@@ -0,0 +1,137 @@
...
+
+        record = {
+            "ts": datetime.now(timezone.utc).isoformat(),
```
stephantul: don't write iso format, just dump the timestamp.

src/semble/stats.py
```diff
@@ -0,0 +1,137 @@
...
+            snippet_chars = record.get("snippet_chars", 0)
+            file_chars = record.get("file_chars", 0)
+            call_type = record.get("call", "search")
+            call_type_counts[call_type] += 1
+            try:
+                record_date = datetime.fromisoformat(record.get("ts", "")).date()
+                in_today = record_date == today
+                in_last_7 = record_date > seven_days_ago
+            except ValueError:
+                in_today = in_last_7 = False  # unparseable timestamp: count in All time only
```
stephantul: how could this happen? AFAIK we write this ourselves. In any case, you should not parse a timestamp to a date and then reparse it again. Just use a timestamp

src/semble/stats.py
```diff
@@ -0,0 +1,137 @@
...
+    lines = [
+        "",
+        "  Semble Token Savings",
+        heavy_line,
+        f"  {'Period':<12}  {'Calls':<6}  Savings",
+        light_line,
+    ]
+    for label, bucket in summary.buckets.items():
+        saved_chars = max(0, bucket.file_chars - bucket.snippet_chars)
+        saved_tokens = saved_chars // 4  # standard ~4 chars/token approximation
```
stephantul, line 117: hmmmm...

src/semble/stats.py
```diff
@@ -0,0 +1,137 @@
...
+        "All time": BucketStats(),
+    }
+    call_type_counts: defaultdict[str, int] = defaultdict(int)
+
+    with path.open() as f:
+        for line in f:
+            try:
+                record = json.loads(line)
+            except json.JSONDecodeError:
+                continue
```
stephantul, line 79: maybe warn here, we don't expect this to happen.

src/semble/stats.py
```diff
@@ -0,0 +1,137 @@
...
+            snippet_chars = record.get("snippet_chars", 0)
+            file_chars = record.get("file_chars", 0)
+            call_type = record.get("call", "search")
```
stephantul: when would we expect any of these to be missing? We write these ourselves.

src/semble/index/index.py
```diff
@@ -216,11 +238,16 @@ def search(
...
+        elif mode == SearchMode.SEMANTIC:
+            results = search_semantic(query, self.model, semantic_index, self.chunks, top_k, selector=selector)
+        elif mode == SearchMode.HYBRID:
+            results = search_hybrid(
                 query, self.model, semantic_index, bm25_index, self.chunks, top_k, alpha=alpha, selector=selector
             )
-        raise ValueError(f"Unknown search mode: {mode!r}")
+        else:
+            raise ValueError(f"Unknown search mode: {mode!r}")
+        if self._file_sizes:
```
stephantul: idem

stephantul (changes requested):
I think most of the code here is overly defensive, lots of guards and checks. But we control the data sources on both ends, so there is likely no need to ever check most of the things. Specifically, I think the gets with defaults of 0 actually hurt rather than help, because they obscure other bugs, and can lead to silent over- or underestimations of the stats.

Issue #81, opened 2026-05-08
Python 3.14 isn't Supported by Dependencies


MrPointer:
## Summary

I just tried semble for the first time. I have `uv` installed and I've installed Python 3.14 with it as it's the latest.
However, installing semble fails due to what seems like a dependency that doesn't yet support Python 3.14 - `tokie`, which is a direct dependency of `chonkie`, which you directly depend on. So it's a transitive dependency for semble basically.

Here's the error log, slightly interleaved with the logs of opencode, the agent I use, and the build attempt made by the dependency:

<details><summary>Error Log</summary>
<p>

```
INFO  2026-05-08T20:49:49 +3887ms service=mcp key=semble mcp stderr:   × Failed to build `tokie==0.0.9`

INFO  2026-05-08T20:49:49 +0ms service=mcp key=semble mcp stderr:   ├─▶ The build backend returned an error
  ╰─▶ Call to `maturin.build_wheel` failed (exit status: 1)

      [stdout]
      Running `maturin pep517 build-wheel -i
      /Users/timorgruber/.cache/uv/builds-v0/.tmp2XxRuI/bin/python
      --compatibility off`

      [stderr]
      🔗 Found pyo3 bindings
      🐍 Found CPython 3.14 at
      /Users/timorgruber/.cache/uv/builds-v0/.tmp2XxRuI/bin/python
      📡 Using build options features from pyproject.toml
      💻 Using `MACOSX_DEPLOYMENT_TARGET=11.0` for aarch64-apple-darwin by
      default
         Compiling target-lexicon v0.13.5
         Compiling syn v2.0.117
         Compiling zmij v1.0.21
         Compiling ring v0.17.14
         Compiling serde_core v1.0.228
         Compiling getrandom v0.3.4
         Compiling zerocopy v0.8.48
         Compiling percent-encoding v2.3.2
         Compiling simd-adler32 v0.3.9
         Compiling adler2 v2.0.1
         Compiling pyo3-build-config v0.24.2
         Compiling serde_json v1.0.149
         Compiling rustls v0.23.37
         Compiling itoa v1.0.18
         Compiling icu_properties_data v1.5.1
         Compiling serde v1.0.228
         Compiling portable-atomic v1.13.1
         Compiling rustls-webpki v0.103.10
         Compiling synstructure v0.13.2
         Compiling pyo3-macros-backend v0.24.2
         Compiling pyo3-ffi v0.24.2
         Compiling miniz_oxide v0.8.9
[...]

Issue #80, opened 2026-05-08
Python version issue when trying to install


phwhalenkeys:
Windows 11 machine with various IT controls adding fun hidden variables :)

I followed the instructions for installing the MCP server into Claude code: I installed uv, then just copy pasted the recommended command into a terminal.

My next claude session reported that the semble MCP had failed. Since I was running in an empty repo, I figured there was some issue at startup. The claude log was unhelpful, just saying that semble timed out after 30000ms.

I tried running "uvx --from "semble[mcp]" semble" in an admin powershell and got the attached log. The biggest thing that stands out is a python incompatibility. It looks like I'm running a newer version of python than something expects under the hood.

[semble log.txt](https://github.com/user-attachments/files/27530776/semble.log.txt)

I'd rather not have to downgrade python (assuming I'm even interpreting this correctly) but open to suggestions.

stephantul:
Hey! Sorry this happened, I'm looking into it. From your logs, this seems to be an issue with the python implementation of `tokie`, which is a transitive dependency for us, via `chonkie`. It could also be an issue with your the rust compiler you have installed.

To be completely fair about it, I think this will be solved by using an older python version. There's no need to downgrade your python installation, you can just set your python version in your environment to python 3.12. It will most likely just work. Unfortunately, I don't know how to do this on Windows.

phwhalenkeys:
I did a quick google based on your suggestion and since I had just installed uv, decided to use it for the problem. I ran 'uv python pin 3.13' in my working directory and then 'uvx --from "semble[mcp]" semble'. This appeared to succeed and bring up the MCP server, so I ctrl-c'ed it and re-ran claude. Semble MCP now resports success!

Hoping this breadcrumb helps a fellow traveler in the future.

PR #84, opened 2026-05-10
feat: add chunking module, remove chonkie

stephantul:
This PR removes [chonkie](https://github.com/chonkie-inc/chonkie) from semble. One of Chonkie's direct dependencies, [tokie](https://github.com/chonkie-inc/tokie), was causing installation failures for users on python 3.14. As `uvx` defaults to the latest stable release of python, users without a pre-existing python environment would pull 3.14 when running, e.g.,

```
uvx --from "semble[mcp]" semble
```

See #81 and #80 for examples of this issue. Upon closer inspection, we don't use any features from either chonkie or tokie in our code, except the code chunker. As the code chunker is mostly a refinement algorithm on top of `tree-sitter`, we opted to reimplement this algorithm.

The actual PR implements a recursive algorithm on top of either:

1) chunks produced by tree-sitter (for languages recognized by us supported by by tree-sitter-language-pack).
2) lines produces by simple splitting (for documentation and languages we don't support).

The algorithm works as follows. Given a hierarchically organized set of chunks, all of which have a start and end span, and a desired chunk size, we create chunks by grouping adjacent chunks if they're below the chunk size. If a chunk is larger than the desired chunk size, we instead split it into its children, and then perform the same algorithm. After this, we then perform a refinement step where we merge adjacent chunks.

For 2), above, we only perform the second step, as lines are not organized hierarchically.

The PR removes 11 dependencies, and maintains 100% test coverage. Performance on our internal benchmark is maintained.


src/semble/chunking/core.py
```diff
@@ -0,0 +1,146 @@
...
+@cache
+def _cached_get_parser(language: SupportedLanguage) -> Parser:
+    """Gets a parser from tree_sitter."""
+    return get_parser(language)
+
+
+def _merge_adjacent_chunks(
+    chunks: list[ChunkBoundary],
+    desired_length: int,
+) -> list[tuple[int, int]]:
```
Pringled: I think this can also return `ChunkBoundary` like `_merge_node_inner`, since the current merged is basically the same thing but just less clearly typed.

src/semble/chunking/core.py
```diff
@@ -0,0 +1,146 @@
...
+            end = child.end_byte
+            length += child_length
+            index += 1
+
+        groups.append(ChunkBoundary(start, end))
+
+    return groups
+
+
+def _merge_node(node: Node, desired_length: int) -> list[tuple[int, int]]:
```
Pringled: Same here, this can return `ChunkBoundary`.

src/semble/chunking/chunking.py
```diff
@@ -0,0 +1,34 @@
+import logging
+
+from semble.chunking.core import chunk, chunk_lines, is_supported_language
+from semble.types import Chunk
+
+logger = logging.getLogger(__name__)
+
+_DESIRED_LENGTH = 1500
```
Pringled: What is this exactly? Could use a small comment to explain what it is and why it's set to this value. Would we ever update this?

Pringled (approved):
Couple of small comments but looks great

PR #99, opened 2026-05-15
feat: add token counts to benchmark

stephantul:
This PR adds token counts to the benchmarks and ablations. The token counts are summarized like the other variables. For each repository, we take the mean number of tokens per retrieved query. This is a good proxy for overall cost, we're not very concerned about outliers.

I refactored some of the benchmarking code, since there was a lot of code duplication. Refactoring the other benchmarks (i.e., the external ones) is another task, and is a bit more involved, so I didn't take that on this time.


Pringled (approved):
LGTM

PR #100, opened 2026-05-17
feat: Refactor ignore

stephantul:
This PR completely refactors the file walker. The code now:

1. respects all `.gitignore`s in the user's directory, not just the root one
2. introduces a `.sembleignore`, which is a `.gitignore` that works in addition to the `.gitignore`, and can be used to override or supplant any `.gitignore` directive
3. adds support for all extensions supported by `tree-sitter-language-pack`.

The `extensions` argument is still supported, and acts like a pseudo `.gitignore`.


src/semble/index/file_walker.py
```diff
@@ -1,149 +1,131 @@
...
-    gitignore = _load_root_gitignore(root)
-    for dirpath, dirnames, filenames in os.walk(root):
-        rel_dir = Path(dirpath).relative_to(root)
-        kept: list[str] = []
-        for dirname in dirnames:
-            if dirname in ignore_dirs:
-                continue
-            if gitignore is not None and _dir_is_gitignored(gitignore, (rel_dir / dirname).as_posix() + "/"):
+    # This should be a list. Traversal is done in order, so the order matters.
+    ignore = []
```
Pringled: Do we keep the `ignore` argument just for the sake of backwards compatibility? Since we just set it to empty here it doesn't actually do anything anymore, so imo we should either remove it or set a deprecation warning and then remove it in 0.2.0.

src/semble/index/file_walker.py
```diff
@@ -1,149 +1,131 @@
...
+    extensions_as_patterns = [f"!*{ext}" for ext in extensions]
+    ignore.extend(extensions_as_patterns)
+    ignore.extend(_DEFAULT_IGNORED_DIRS)
+    # Always give user patterns preference
+    ignore.extend(ignore or [])
```
Pringled: Related to my comment above, this never happens since ignore is set to an empty list.

src/semble/index/create.py
```diff
@@ -27,18 +28,15 @@ def create_index_from_path(
...
     :raises ValueError: if no items were found, no index can be created.
     :return: A bm25 index, vicinity index and list of chunks
     """
-    extensions = filter_extensions(extensions, include_text_files=include_text_files)
-
     chunks: list[Chunk] = []
-
-    for file_path in walk_files(path, extensions, ignore):
-        language = language_for_path(file_path)
+    extensions = get_extensions(include_text_files, extensions)
```
Pringled, line 37: I think there's a behavioral change with this new logic where if a user supplies extensions, they always get those in addition to our defaults, while previously we _only_ indexed files of that extension. TBH I don't have a preference, I think both are sensible, but do we want to still allow support for only indexing a certain type of file?

src/semble/index/file_walker.py
```diff
@@ -1,149 +1,131 @@
...
-            yield file_path
+
+            if pattern.match_file(relative_str) is not None:
+                ignored = pattern.include
+
+    return ignored
+
+
+def _walk(
+    dir: Path,
```
Pringled: nitpick but this shadows the dir function

Pringled (approved):
Couple of small things, but this is a big improvement overall

Issue #78, opened 2026-05-08
First run stuck


usuf13:
## Environment

* OS: Windows 10
* VS Code: v1.119.0
* Claude Code for VS Code: v2.1.133

## Installation

Installed via:

```bash
claude mcp add semble -s user -- uvx --from "semble[mcp]" semble
```

## Problem

The very first session call always gets stuck indefinitely.
I waited up to 20 minutes one time, but the operation never completed.

After manually interrupting the first run, all subsequent calls work normally.

I also reproduced the same behavior when running Claude in the terminal (using debug mode).
No obvious errors appear in the debug logs.

Additionally, the index file is never created inside the project folder.

## Logs

```text
2026-05-08T08:51:00.817Z [DEBUG] [FileIndex] getProjectFiles called, respectGitignore=true
2026-05-08T08:51:00.818Z [DEBUG] [FileIndex] getFilesUsingGit called
2026-05-08T08:51:00.822Z [DEBUG] Fast mode unavailable: Fast mode requires extra usage billing · /extra-usage to enable
2026-05-08T08:51:00.824Z [DEBUG] High write ratio: blit=0, write=1078 (100.0% writes), screen=22x142
2026-05-08T08:51:00.828Z [DEBUG] AutoUpdaterWrapper: Installation type: native
2026-05-08T08:51:00.830Z [DEBUG] Fast mode unavailable: Fast mode requires extra usage billing · /extra-usage to enable
2026-05-08T08:51:00.861Z [DEBUG] [FileIndex] git ls-files (tracked) took 43ms
2026-05-08T08:51:00.861Z [DEBUG] [FileIndex] git ls-files: 2604 tracked files in 43ms
2026-05-08T08:51:00.864Z [DEBUG] [FileIndex] using git ls-files result (2604 files)
2026-05-08T08:51:00.865Z [DEBUG] [FileIndex] skipped index rebuild — tracked paths unchanged
2026-05-08T08:51:00.865Z [DEBUG] [FileIndex] cache refresh completed in 48ms
2026-05-08T08:51:00.924Z [DEBUG] [FileIndex] background untracked fetch: 10 files
2026-05-08T08:51:00.925Z [DEBUG] [FileIndex] skipped index rebuild — merged paths unchanged
2026-05-08T08:51:00.991Z [DEBUG] Fast mode unavailable: Fast mode requires extra usage billing · /extra-usage to enable
[...]

PR #101, opened 2026-05-17
feat: disable multi processing

stephantul:
This PR disables multiprocessing. This makes indexing slightly slower (about 8% on average on our benchmark). However, we believe that #78 is caused by multiprocessing running in an async loop. Given that we don't have a good way to reproduce this issue on Windows machines (it does not happen in CI), this is a safe way to go about it.

This PR also fixes some typing stuff, where we were assuming that the dimensionality of our model was always fixed to 256. This is not true, and would (maybe) block our users from using custom models.

### DIFF
--- a/src/semble/index/dense.py
+++ b/src/semble/index/dense.py
@@ -27,8 +27,8 @@ def load_model(model_path: str | None = None) -> Encoder:
 def embed_chunks(model: Encoder, chunks: list[Chunk]) -> npt.NDArray[np.float32]:
     """Embed chunks using the configured model."""
     if not chunks:
-        return np.empty((0, 256), dtype=np.float32)
-    return np.array(model.encode([c.content for c in chunks]), dtype=np.float32)
+        return np.empty((0, model.dim), dtype=np.float32)
+    return np.array(model.encode([c.content for c in chunks], use_multiprocessing=False), dtype=np.float32)
 
 
 class SelectableBasicBackend(CosineBasicBackend):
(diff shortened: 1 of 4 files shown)


src/semble/types.py
```diff
@@ -27,7 +27,12 @@ class CallType(str, Enum):
 class Encoder(Protocol):
     """Protocol for embedding models."""
 
-    def encode(self, texts: Sequence[str], /) -> EmbeddingMatrix:
+    @property
+    def dim(self) -> int:
```
stephantul, line 31: Adding this kind of makes this protocol useless, since sentence-transformers don't have `dim`. So, in effect, this makes `Encoder` equivalent to `StaticModel`. Apart from wrapping either, I don't see a nice way around this. I think this is fine for now, however.

src/semble/index/dense.py
```diff
@@ -27,8 +27,8 @@ def load_model(model_path: str | None = None) -> Encoder:
 def embed_chunks(model: Encoder, chunks: list[Chunk]) -> npt.NDArray[np.float32]:
     """Embed chunks using the configured model."""
     if not chunks:
-        return np.empty((0, 256), dtype=np.float32)
-    return np.array(model.encode([c.content for c in chunks]), dtype=np.float32)
+        return np.empty((0, model.dim), dtype=np.float32)
+    return np.array(model.encode([c.content for c in chunks], use_multiprocessing=False), dtype=np.float32)
```
stephantul, line 31: This line de facto makes this unusable with a non-static model. We could check whether it is a StaticModel etc. wdyt?

PR #103, opened 2026-05-17
fix: determinism in file order

stephantul:
This PR fixes a low-key determinism bug. Due to python's HASHSEED behavior, it was possible that two chunks from the same file with the same score would get a different tie break. Because of the way we calculate boosting, this would result in different rankings, and slightly different NDCG@10 scors.

The fix is just sorting the chunks, leading to a stable key which does not rely on a specific hash seed.

### DIFF
--- a/src/semble/index/file_walker.py
+++ b/src/semble/index/file_walker.py
@@ -65,7 +65,7 @@ def walk_files(root: Path, extensions: Sequence[str], ignore: Sequence[str] | No
     ignored = []
     extensions_as_patterns = [f"!*{ext}" for ext in extensions]
     ignored.extend(extensions_as_patterns)
-    ignored.extend(_DEFAULT_IGNORED_DIRS)
+    ignored.extend(sorted(_DEFAULT_IGNORED_DIRS))
     # Always give user patterns preference
     ignored.extend(ignore or [])
     base_spec = GitIgnoreSpec.from_lines(ignored, backend="simple")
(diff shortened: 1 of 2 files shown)


src/semble/index/file_walker.py
```diff
@@ -65,7 +65,7 @@ def walk_files(root: Path, extensions: Sequence[str], ignore: Sequence[str] | No
     ignored = []
     extensions_as_patterns = [f"!*{ext}" for ext in extensions]
     ignored.extend(extensions_as_patterns)
-    ignored.extend(_DEFAULT_IGNORED_DIRS)
```
stephantul, line 68: This was also a possible source of errors, although I think it actually wasn't causing an issue.

src/semble/index/file_walker.py
```diff
@@ -109,23 +109,21 @@ def _walk(
...
     spec = _load_ignore_for_dir(directory)
     if spec is not None:
-        active_specs = [
+        inherited_specs = [
             *inherited_specs,
             IgnoreSpec(base=directory, spec=spec),
         ]
 
-    for item in directory.iterdir():
+    for item in sorted(directory.iterdir()):
```
stephantul, line 119: idem, this could result in issues, so just to rule it out, we sort here.

src/semble/index/file_walker.py
```diff
@@ -109,23 +109,21 @@ def _walk(
     inherited_specs: list[IgnoreSpec],
 ) -> Iterator[Path]:
     """Recursive function for walking files under a directory."""
-    active_specs = inherited_specs
-
     spec = _load_ignore_for_dir(directory)
     if spec is not None:
-        active_specs = [
+        inherited_specs = [
```
stephantul, line 114: this is done like this to avoid an in-place edit of inherited_specs.

src/semble/search.py
```diff
@@ -107,10 +107,16 @@ def search_hybrid(
     normalized_semantic = _rrf_scores(semantic_scores)
     normalized_bm25 = _rrf_scores(bm25_scores)
 
+    # Sort by the file path and start line to
+    # counteract randomness introduces by hashing.
+    all_candidates = sorted(
+        {*normalized_semantic, *normalized_bm25},
```
stephantul, line 113: this is the actual bugfix

Pringled (approved):

PR #110, opened 2026-05-18
remove invisible spaces

nilsherzig:
idk what happened there, feel free to just fix that yourself (or ignore it haha). But there used to be https://unicode-explorer.com/c/200B zero with spaces for some reason


Pringled:
Hey @nilsherzig, these are actually needed, they're there to prevent the inner triple-backticks from terminating the outer code block. See this for example on your branch:
image

PR #120, opened 2026-05-20
feat: add windsurf and zed targets to semble init

MackDing:
## Summary\n- add  and  as supported  targets\n- write agent files to the documented paths for each tool\n- document the new init commands in README\n- add focused tests for the new output paths\n\n## Why\nThe README already documents MCP setup for Windsurf and Zed, but the dedicated sub-agent setup section did not support them via . This change keeps the CLI aligned with the docs and extends an existing workflow in a small, low-risk way.\n\n## Verification\n- added targeted tests for the new agent output paths in \n- local test execution was partially blocked by missing optional dependency  in the current environment, so I verified the changed path logic directly and kept the change surface small

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,7 +327,7 @@ If `semble` is not on `$PATH`, use `uvx --from "semble[mcp]" semble` in its plac
 
 ### Sub-agent setup
 
-Claude Code, Gemini CLI, Cursor, OpenCode, GitHub Copilot CLI, and Kiro all support a dedicated semble search sub-agent. Run `semble init` once in your project root:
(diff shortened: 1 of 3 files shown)


Pringled:
@MackDing neither windsurf nor zed supports an agents dir with custom agents from what I can tell. Also this is not in line with our contributing guidelines, feel free to open an issue.

MackDing:
Thanks for the quick review — that makes sense.

I based this on the README platform coverage, but if Windsurf/Zed don't support custom agents directories and this should go through an issue first, I'm happy to defer to that process.

I'll close this PR to avoid noise. Thanks again.

PR #136, opened 2026-05-21
feat: Make model loading non-blocking in MCP serve

### LINKED ISSUE #133: MCP server semble_search tool times out on first queries (CLI works fine)
## Environment
- **Semble version:** v0.1.10
- **OS:** macOS (darwin)
- **Agent:** OpenCode (Claude Code-compatible MCP client)
- **Python/uv:** uvx via `uvx --from "semble[mcp]" semble`
- **Repo size:** Java project, ~48 relevant files for JDBC module
## Description
When using the Semble MCP server through an agent harness (OpenCode), the `semble_search` tool consistently returns `MCP error -32001: Request timed out` on the first several queries. The exact same queries work immediately via the CLI.
## Reproduction Steps
1. Configure Semble as MCP server in the agent config:
   ```json
   {
     "mcp": {
       "semble": {
         "type": "local",
         "command": ["uvx", "--from", "semble[mcp]", "semble"]
       }
     }
   }
2. Start agent session and invoke semble_search with any query:
- Query: JDBC type conversion ResultSet PreparedStatement DataType
- Mode: hybrid or bm25
- top_k: 1–5
- repo: <path-to-java-project>
3. Result: MCP error -32001: Request timed out (repeated across 5+ attempts)
4. Run the same query via CLI:
uvx --from "semble[mcp]" semble search "AbstractJdbcRowDataConverter" <path-to-java-project>
Result: Returns results in ~1 second.
Expected Behavior
MCP tool should return search results within a reasonable timeout, matching CLI performance.
Actual Behavior
MCP tool times out with MCP error -32001: Request timed out on every attempt. CLI works fine with same repo and query.
Additional Context
- CLI version works correctly after a manual semble search run
- No ~/.cache/semble directory exists (first run)
- MCP server appears to start successfully (no startup errors)
- Timeout occurs even with minimal query (DataType, bm25 mode, top_k: 1)
- After CLI indexing, MCP tool eventually works on subsequent calls
- Suggests the MCP server may not be handling initial indexing asynchronously or may have a different timeout/initialization path than CLI
Possible Causes
1. MCP server initialization blocks on indexing without proper async handling
2. MCP transport layer timeout is too short for cold-start indexing
3. Index cache path differs between CLI and MCP server (~/.cache/semble not created)
4. MCP server may not be reusing the same index location as CLI
Workaround
Run semble search via CLI once to pre-build the index, then MCP tool works on subsequent calls.

Pringled:
This PR makes the MCP's serve non-blocking for model loading (as well as direct indexing from a given path). This should hopefully resolve https://github.com/MinishLab/semble/issues/133, or at least help since the model download no longer blocks the initial startup for Semble. `force_download` is also set to False for the model so we don't check HF everytime if the model is already cached. Tested locally by deleting my own cached model and it makes the server respond in ~0.3 seconds instead of ~5 seconds (on very very fast internet).

### DIFF
--- a/src/semble/mcp.py
+++ b/src/semble/mcp.py
@@ -113,27 +113,70 @@ def create_server(cache: _IndexCache, default_source: str | None = None) -> Fast
 
 
 async def serve(path: str | None = None, ref: str | None = None, include_text_files: bool = False) -> None:
-    """Start an MCP stdio server, optionally pre-indexing a default source."""
-    model = await asyncio.to_thread(load_model)
-    cache = _IndexCache(model=model, include_text_files=include_text_files)
-    if path:
-        await cache.get(path, ref=ref)
-        if not _is_git_url(path):
-            await cache.start_watcher(path)
+    """Start an MCP stdio server, optionally pre-indexing a default source.
 
+    The MCP transport opens before the embedding model finishes loading, so the
+    ``initialize`` / ``tools/list`` handshake responds even on a cold HF cache.
+    Tool calls await the model implicitly via :class:`_IndexCache`.
+    """
+    cache = _IndexCache(include_text_files=include_text_files)
+
+    async def _load_and_prewarm() -> None:
+        try:
+            model = await asyncio.to_thread(load_model)
+        except Exception as exc:
+            logger.exception("Failed to load embedding model")
+            cache.set_model_error(exc)
+            return
+        cache.set_model(model)
+        if path:
+            try:
+                await cache.get(path, ref=ref)
+            except Exception:
+                logger.warning("Failed to pre-index %r at startup", path, exc_info=True)
+            if not _is_git_url(path):
+                await cache.start_watcher(path)
+
+    init_task = asyncio.create_task(_load_and_prewarm())
     server = create_server(cache, default_source=path)
-    await server.run_stdio_async()
+    try:
+        await server.run_stdio_async()
+    finally:
+        if not init_task.done():
+            init_task.cancel()
 
 
 class _IndexCache:
     """Cache of indexed repos and local paths for the lifetime of the MCP server process."""
 
-    def __init__(self, model: Encoder, include_text_files: bool = False) -> None:
-        """Initialise an empty cache with a shared embedding model."""
-        self._model = model
+    def __init__(self, model: Encoder | None = None, include_text_files: bool = False) -> None:
+        """Initialise an empty cache; ``model`` may be supplied later via :meth:`set_model`."""
+        self._model: Encoder | None = model
+        self._model_error: BaseException | None = None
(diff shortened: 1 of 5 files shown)


src/semble/mcp.py
```diff
@@ -114,27 +114,57 @@ async def find_related(
...
-    def __init__(self, model: Encoder, include_text_files: bool = False) -> None:
-        """Initialise an empty cache with a shared embedding model."""
-        self._model = model
+    def __init__(self, model: Encoder | None = None, include_text_files: bool = False) -> None:
+        """Initialise an empty cache."""
+        self._model: Encoder | None = model
+        self._model_error: BaseException | None = None
+        self._model_ready = asyncio.Event()
+        if model is not None:
+            self._model_ready.set()
```
stephantul, line 155: can't this be a property?

src/semble/mcp.py
```diff
@@ -114,27 +114,57 @@ async def find_related(
 
 async def serve(path: str | None = None, ref: str | None = None, include_text_files: bool = False) -> None:
     """Start an MCP stdio server, optionally pre-indexing a default source."""
-    model = await asyncio.to_thread(load_model)
-    cache = _IndexCache(model=model, include_text_files=include_text_files)
-    if path:
-        await cache.get(path, ref=ref)
-        if not _is_git_url(path):
-            await cache.start_watcher(path)
+    cache = _IndexCache(include_text_files=include_text_files)
```
stephantul, line 117: why not load the model here, and just pass in the model as an argument to `_IndexCache`? That saves you from the whole `_model_ready` logic. The model loading can still be done async (although not sure why)

src/semble/mcp.py
stephantul: Also see comment above, maybe just leave it out?

stephantul (approved):
Approval contingent on discussion, maybe i misunderstood

src/semble/mcp.py
Pringled: _model_ready is an asyncio.Event so it can't be a property I think

src/semble/mcp.py
Pringled: That would not solve the existing issue since the model loading would still be on the blocking path, with this logic we can immediately start the server (and respond to the agent that it's healthy) and the model downloads in the background so it only blocks the first search (which should have a larger timeout). E.g. here's an example if I use your suggestion:

|            | Before | After |
|------------|--------|-------|
| initialize | 3.95s  | 0.57s |
| tools/list | 3.95s  | 0.57s |

My suspicion is that the tool listing could be the blocking part, we basically don't show to the agent what tools are available until the server is ready but the server can potentially take long to start if the user has slow internet or if HF responds slowly for some reason.

PR #140, opened 2026-05-21
feat: add index persistence

stephantul:
This PR adds index persistence for use in the CLI, which is a feature requested by several users.

This is a big PR! Sorry for that. This PR:

* Adds a new `index` command to the CLI.

You can now index a repository as follows:

```bash
semble index -o "my_index"
```

and then search using:

```bash
semble search "where is persistency defined?" --index "my_index"
```

This greatly speeds up subsequent searches. Loading a decently-sized index takes 200-400ms.

* Adds persistence to the embedding backend.  This was necessary because we override the basicbackend a little weirdly. Something we can think about doing differently
* Adds persistence to the index itself. All components are saved in subfolders, except the model. For the model, we save the _name_ of the model. To facilitate saving itself, I added helpers to `Chunk`. These are thin wrappers around `asdict` and a dictionary expansion.
* Removed the `Encoder` protocol: this no longer made sense because we use the saving and loading methods in `model2vec`.

The ugly part of this is that I chose to refactor a large part of the code: we now now longer pass a `model` to the index when building it. Instead I use a path to the model. This path is then used to load the model, and reverts to the default model when `None`. This is a more elegant construction I think, since this allows us to store the model path, and also cache model loading more efficiently.

Follow-up tasks:
* Not all benchmarks work, but the most important ones (i.e., the regular one and ablations) still work.


src/semble/index/index.py
```diff
@@ -237,3 +245,46 @@ def search(
...
+
+        chunks = []
+        for chunk_item in chunk_data:
+            chunks.append(Chunk.from_dict(chunk_item))
+        root_path = metadata["root_path"]
+        model_path = metadata["model_path"]
+        if root_path:
+            root_path = Path(root_path)
+
+        model = StaticModel.from_pretrained(model_path)
```
Pringled: Should this not just call `load_model`?  So you disable the progressbars and set force download?

src/semble/index/index.py
```diff
@@ -237,3 +245,46 @@ def search(
...
+    @classmethod
+    def load_from_disk(cls: type[SembleIndex], path: Path | str) -> SembleIndex:
+        """Load the index from disk."""
+        path = Path(path)
+        if not path.exists():
+            raise FileNotFoundError(f"Index not found at {path}")
+        persistence_paths = PersistencePath.from_path(path)
+        bm_25_index = BM25.load(persistence_paths.bm25_index)
+        semantic_index = SelectableBasicBackend.load(persistence_paths.semantic_index)
+        metadata = orjson.loads(open(persistence_paths.metadata).read())
```
Pringled: I guess `with open()` is a slightly safer pattern but I've honestly never seen any issues in practice with this so yeah... whatever you want

src/semble/index/index.py
```diff
@@ -237,3 +245,46 @@ def search(
         )
         save_search_stats(results, CallType.SEARCH, self._file_sizes)
         return results
+
+    @classmethod
+    def load_from_disk(cls: type[SembleIndex], path: Path | str) -> SembleIndex:
+        """Load the index from disk."""
+        path = Path(path)
+        if not path.exists():
```
Pringled, line 283: Hmm is this enough to check? If you use the wrong path here that does exist (but without the data) you get a nasty error when it tries to load the stuff later. Then again checking it adds a bunch of ifs iguess

src/semble/cli.py
```diff
@@ -66,6 +66,13 @@ def _mcp_main() -> None:
     asyncio.run(serve(args.path, ref=args.ref, include_text_files=args.include_text_files))
 
 
+def _run_index(*, path: str, include_text_files: bool = False, out: str) -> None:
```
Pringled, line 85: Do we intentionally only support local paths or do we also want to do git urls? If we don't want to support it I think we should maybe document that or raise an error here if someone does try to do a git url

src/semble/index/index.py
```diff
@@ -89,21 +96,21 @@ def stats(self) -> IndexStats:
     def from_path(
         cls,
         path: str | Path,
-        model: Encoder | None = None,
         extensions: Sequence[str] | None = None,
         include_text_files: bool = False,
+        model_path: str | None = None,
```
Pringled, line 126: Technically a breaking API change but I don't think anyone uses Semble as a programmatic API tbh wdyt?

src/semble/mcp.py
```diff
@@ -119,7 +118,7 @@ async def serve(path: str | None = None, ref: str | None = None, include_text_fi
     async def _load_and_prewarm() -> None:
         """Pre-load the model and optionally pre-index the default source in parallel with starting the server."""
         try:
-            cache._model = await asyncio.to_thread(load_model)
+            _, cache._model_path = await asyncio.to_thread(load_model)
```
Pringled, line 127: `load_model()` here caches on (), but later `_IndexCache.get` calls `load_model("minishlab/potion-code-16M")` via from_path so it won't hit the cache, I think we can just use None everywhere, should be safe and then we do have a working cache?

Pringled (approved):
Very very nice, some minor things

src/semble/index/index.py
stephantul: Added

PR #141, opened 2026-05-21
fix(mcp): honor --ref when using the default pre-indexed repo

LucasCavalheri:
## Summary

Fix an MCP cache inconsistency when the server is started with a git repository URL and `--ref`, but subsequent `search` / `find_related` tool calls omit `repo`.

Previously:

* startup prewarming indexed the repository using a cache key like `url@ref`
* tool calls without `repo` resolved cache entries using only `url`
* this caused cache misses and unnecessary re-cloning/re-indexing of the repository default branch

This change ensures that tool calls targeting the default startup repository consistently reuse the same pre-warmed cache entry.

---

### Ref-aware default source resolution

Added `_resolve_ref()` behavior so git URLs use `default_ref` when:

* `repo is None`
* `repo == default_source`

This allows MCP tools to correctly reuse the startup repository reference.

---

### Ref-aware cache lookups

`_get_index()` now resolves and forwards the effective `ref` into:

```python
cache.get(source, ref=ref)
```

instead of always using only the source URL.

---

### Propagate default ref through server startup

`create_server()` now accepts:

```python
default_ref: str | None
```

and `serve()` forwards the startup `--ref` value into MCP tool resolution.

---

### Ref-aware eviction

`evict()` now accepts an optional `ref` argument so git cache keys such as:

```text
url@ref
```

are correctly removed during failures or cancellations.

This prevents orphaned cache entries.

---

## Test Plan

* [x] `uv run pytest tests/test_mcp.py -q`
* [x] `uv run ruff check src/semble/mcp.py tests/test_mcp.py`
* [x] `uv run mypy src/semble/mcp.py`

### Added Tests

* `test_index_cache_git_ref_key`
* `test_get_index_reuses_default_ref`
* `test_tool_search_reuses_default_ref_cache`
* `test_cache_evict_with_ref`

---

## Behavior Before

```text
startup:
  cache key -> url@ref

tool call without repo:
  cache lookup -> url

=> cache miss
=> second clone/index on default branch
```

---

## Behavior After

```text
startup:
  cache key -> url@ref

tool call without repo:
  cache lookup -> url@ref

=> cache reuse
=> no unnecessary re-index
```

---

## Compatibility

No public API changes to `SembleIndex`.

No behavioral changes for:

* local paths
* non-git repositories
* existing MCP tool usage without `--ref`

This only affects internal cache resolution for default git startup repositories.

---

## Manual Verification

```bash
semble https://github.com/MinishLab/semble --ref <tag-or-branch>
```

Then invoke MCP `search` / `find_related` without `repo`.

Expected behavior:

* no second clone
* no second indexing pass
* same pre-warmed cache entry reused

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ Add to `~/.kiro/settings/mcp.json` (or `.kiro/settings/mcp.json` in your project
 <details>
 <summary>Zed</summary>
 
-Add to `~/.config/zed/settings.json` (or `.zed/settings.json` in your project):
+**Via settings** — add to `~/.config/zed/settings.json` (or `.zed/settings.json` in your project):
 
 ```json
 {
@@ -273,6 +273,20 @@ Add to `~/.config/zed/settings.json` (or `.zed/settings.json` in your project):
       "command": "uvx",
       "args": ["--from", "semble[mcp]", "semble"]
     }
+  }
+}
+```
+
+**Via UI** — in the Agent Panel:
+
+1. Open **MCP Servers** → **Add Custom Server...** → **Local**
+2. Paste:
+
+```json
+{
+  "semble": {
+    "command": "uvx",
+    "args": ["--from", "semble[mcp]", "semble", "mcp"]
   }
 }
 ```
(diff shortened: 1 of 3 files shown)


Pringled:
@LucasCavalheri see https://github.com/MinishLab/semble/blob/main/CONTRIBUTING.md.

PR #142, opened 2026-05-21
feat: Add --content flag for docs, config, and code search

Pringled:
This PR adds a `content` parameter to control what gets indexed: code (default), docs, config, or all. The prompts are also updated so they are aware of the new parameter. Example use:

CLI:
```
semble search "deployment guide" ./my-project --content docs
semble search "authentication" ./my-project --content all
```

Python:
```python
index = SembleIndex.from_path("./my-project", content=ContentType.DOCS)
```
This also deprecates `include_text_files` which is equal to content = all. It still exists, but just gets overridden by content = all and a warning is shown. 

The language sets are refactored into _DOC_LANGUAGES, _CONFIG_LANGUAGES, and _CODE_LANGUAGES to make it a bit clearer.


src/semble/index/files.py
```diff
@@ -447,12 +452,15 @@ def detect_language(file_name: Path) -> str | None:
...
 
-def get_extensions(include_text_files: bool, extensions: Sequence[str] | None) -> list[str]:
+def get_extensions(content: ContentType, extensions: Sequence[str] | None) -> list[str]:
     """Returns a list of supported file extensions."""
-    if include_text_files:
+    languages: set[str] | frozenset[str]
+    if content == ContentType.ALL:
         languages = ALL_LANGUAGES
+    elif content == ContentType.DOCS:
+        languages = _DOC_LANGUAGES
```
stephantul: one thing I find a little bit odd here is that it is apparently not possible for users to select `Config`, even though it is present as a category. I understand the reasoning, you don't necessarily want to expose this to the user and maintain feature parity with `include_text_files`. Nevertheless, we do offer users to search through `config`, but only by passing `all`. This seems a bit confusing.

E.g., a user does a search with "all" sees something they want, then they want to narrow, but there's actually no option to narrow it down at all. 

So, maybe as a middle ground, we could add it, but not advertise it? Or maybe add it internally, and not expose the `config` as a `cli` arg?

stephantul (approved):

PR #147, opened 2026-05-22
feat: add JSON output

### LINKED ISSUE #127: suggestion: add a flag for json output
On top of having agents use it, I've been using semble myself a lot. I use emacs heavily and wanted to write a package to search repos using semble. I know I could just use a little python script that imports semble, but I think having json output built-in would be really convenient for anyone who wants to integrate semble into something.

stephantul:
This PR adds JSON output as the default output format. The actual output format is also a little bit different: we have a top-level "query" key and "results" key. This can be parsed by `jq`. 

e.g.,

```bash
semble search "how do I save a model" | jq '.results[].chunk.file_path'
```

gives you a list of all file paths. Fixes #127

### DIFF
--- a/src/semble/cli.py
+++ b/src/semble/cli.py
@@ -1,5 +1,6 @@
 import argparse
 import asyncio
+import json
 import sys
 import warnings
 from enum import Enum
@@ -185,9 +186,10 @@ def _cli_main() -> None:
     if args.command == "search":
         results = index.search(args.query, top_k=args.top_k)
         if not results:
-            print("No results found.")
+            out = {"error": "No results found."}
         else:
-            print(format_results(f"Search results for: {args.query!r}", results))
+            out = format_results(args.query, results)
+        print(json.dumps(out, indent=4))
 
     elif args.command == "find-related":
         chunk = resolve_chunk(index.chunks, args.file_path, args.line)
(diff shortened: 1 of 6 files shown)


src/semble/mcp.py
```diff
@@ -77,8 +78,8 @@ async def search(
             return str(exc)
         results = index.search(query, top_k=top_k)
         if not results:
-            return "No results found."
-        return format_results(f"Search results for: {query!r}", results)
+            return json.dumps({"error": "No results found."})
+        return json.dumps(format_results(query, results), indent=4)
```
Pringled: Does indentation actually matter for an agent? It's just extra tokens afaik so maybe it's better to just do a json output with the least amount of formatting for the MCP case?

src/semble/cli.py
```diff
@@ -196,6 +198,7 @@ def _cli_main() -> None:
             sys.exit(1)
         results = index.find_related(chunk, top_k=args.top_k)
         if not results:
-            print(f"No related chunks found for {args.file_path}:{args.line}.")
+            out = {"error": "No related chunks found."}
```
Pringled: The MCP has a slightly different (and IMO more informative) error message

Pringled (approved):
LGTM two minor concerns, and do we want to support plain text output still via a flag or just drop it entirely?

Issue #149, opened 2026-05-23
Semble MCP fails without explicit `repo` while CLI falls back to cwd


docevilOck:
# Issue Draft[ BUG ]: Semble MCP fails without explicit `repo` while CLI falls back to cwd

## Summary

In the current Codex + Semble setup, `Semble MCP` search fails when the caller does not pass `repo`, even if the current working directory is already the target repository.

The same machine and same repository succeed with the Semble CLI, because the CLI uses the current directory as its default path.

This creates an inconsistent behavior:

- `semble search ...` works in the repo root
- `mcp__semble__.search(...)` fails unless `repo` is passed explicitly or the MCP server was started with a default source

## Impact

- Agents can incorrectly conclude that "Semble does not work in this repository"
- Search reliability depends on call style rather than repository state
- Any workflow that assumes MCP will inherit cwd is broken

## Environment

- OS: Windows 11
- Shell: PowerShell 7.6.0
- Repository: `E:\Repo\opencode-dynamic-context-pruning`
- Codex MCP config:

```toml
[mcp_servers.semble]
command = "semble"
```

This means the MCP server is started without a default repository path.

### Failing MCP call

```text
mcp__semble__.search(query="compress pipeline")
```

Actual result:

```text
No repo specified and no default index. Pass an https:// or http:// git URL or local directory path as `repo`.
```

### Working CLI call

Run in:

```text
E:\Repo\opencode-dynamic-context-pruning
```

Command:

```powershell
semble search "compress pipeline"
```

Actual result:

- Returns normal search results from this repository

### Working MCP call with explicit repo

```text
mcp__semble__.search(
  query="compress pipeline",
  repo="E:\\Repo\\opencode-dynamic-context-pruning"
)
```

Actual result:

- Returns normal search results from this repository

## Expected Behavior

One of the following should be true:

1. MCP should behave like CLI and use the current repository context by default
2. Or the integration must clearly require callers to always pass `repo`

At minimum, agent-side integrations should not assume cwd inheritance.

## Actual Behavior

MCP does not infer the current repository from cwd.

If the MCP server is started without a default source and the caller omits `repo`, the request fails immediately.

## Root Cause

This is caused by an implementation difference between Semble CLI and Semble MCP.

### CLI behavior

In local Semble installation, CLI search defines:

- `path` default = `.`

So `semble search "query"` uses the current working directory automatically.

### MCP behavior

In local Semble installation, MCP resolves the index source as:

```text
source = repo or default_source
```

[...]

PR #150, opened 2026-05-23
fix: default MCP search to server cwd

docevilOck:
fix the issue:Semble MCP fails without explicit repo while CLI falls back to cwd #149
## Summary

This PR makes the MCP server match the CLI's local default behavior. When the MCP server starts without an explicit path, it now uses the server process current working directory as the default local repo. MCP tool calls can omit `repo` for that default local source, while explicit `repo` values still take precedence for cross-repo or remote searches.

## Problem

Before this change, local code search behaved differently between CLI and MCP:

- CLI commands defaulted to the current directory for local searches.
- MCP tool calls failed when both `repo` and startup `default_source` were absent.

That meant an agent running inside a local repository could call the MCP `search` tool without `repo` and unexpectedly get a "No repo specified" error, even though the equivalent CLI workflow worked from the same directory.

## Root Cause

The MCP server only passed through the optional startup `path` as `default_source`. If the server was started without that path, `default_source` stayed empty. Later, tool calls resolved the index source from `repo or default_source`; when both were empty, the call failed.

The gap was not in indexing or search itself. The failure condition was specifically triggered by MCP calls that omitted `repo` while the MCP server had also been started without an explicit default path.

## Changes

- Added centralized default-source resolution in `src/semble/mcp.py`.
- When `serve()` receives no startup path, it resolves `Path.cwd()` and uses that as the MCP default local source.
- Reused the resolved default source for pre-indexing, file watching, and server tool defaults.
- Preserved explicit `repo` precedence, so remote repos and cross-repo local searches still work by passing `repo`.
- Updated MCP instructions and README text to document the new default behavior.
- Added tests covering omitted `repo`, explicit `repo` precedence, and no-path server startup behavior.

## Behavior After This PR

- `mcp search(query=...)` uses the MCP server working directory when no `repo` is passed.
- `mcp search(query=..., repo="E:/Repo/other")` searches the explicit local repo.
- `mcp search(query=..., repo="https://github.com/org/repo")` searches the explicit remote repo.
- Explicit `repo` remains the recommended pattern for multi-repo or remote-repo workflows.

## Verification

- `uv run pytest tests/test_cli.py tests/test_mcp.py -q`
  - Result: passed, `75 passed`
- `uv run ruff check src/ tests/`
  - Result: passed
- `uv run pytest -q`
  - Result: `184 passed, 1 failed`
  - The remaining failure is `tests/test_file_walker.py::test_walk_files_skips_symlinks`, caused by Windows symlink privilege error `WinError 1314`. This is unrelated to the MCP default-repo change.

## Risk

The main behavior change is that MCP servers started without a path now default to their process working directory instead of having no default source. This aligns MCP with the CLI for local repository usage, but agent harnesses should still pass `repo` explicitly when they need deterministic cross-repo behavior.

## Commit

`6468f8c fix: default MCP search to server cwd`

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ Add Semble to Claude Code (requires [uv](https://docs.astral.sh/uv/getting-start
 claude mcp add semble -s user -- uvx --from "semble[mcp]" semble
 ```
 
+When started without a path, the MCP server uses its current working directory as the default local repo. Tool calls can omit `repo` for that default, or pass an explicit local path / `https://` URL to search another repo.
+
 Using another agent harness? See [MCP Server](#mcp-server) below for per-agent setup.
 
 ### Bash / AGENTS.md
@@ -134,6 +136,8 @@ uv cache clean semble          # for MCP users (restart your MCP client after)
 ## MCP Server
 
 Semble can run as an MCP server so agents can search any codebase directly. Repos are cloned and indexed on demand, and indexes are cached for the lifetime of the session. Local paths are watched for file changes and re-indexed automatically.
+
+If the MCP server starts without a path argument, its current working directory becomes the default local source. Calls to `search` and `find_related` may omit `repo` to use that default. Pass `repo` explicitly when searching another local repo or any remote repo.
 
 ### Setup
 
--- a/src/semble/mcp.py
+++ b/src/semble/mcp.py
@@ -21,7 +21,7 @@ logger = logging.getLogger(__name__)
 
 _REPO_DESCRIPTION = (
     "https:// or http:// git URL (e.g. https://github.com/org/repo) or local directory path to index and search. "
-    "Required when no default index was configured at startup. "
+    "Defaults to the MCP server's startup path, or its current working directory if no startup path was provided. "
     "The index is cached after the first call, so repeat queries are fast."
 )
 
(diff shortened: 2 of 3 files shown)


Pringled:
@docevilOck see https://github.com/MinishLab/semble/blob/main/CONTRIBUTING.md.

docevilOck:
> @docevilOck see https://github.com/MinishLab/semble/blob/main/CONTRIBUTING.md.

ok，get it

docevilOck:
> @docevilOck see https://github.com/MinishLab/semble/blob/main/CONTRIBUTING.md.
this rule said if want add a feature, should discuss with you, how could i access you, just in the issue communicate that?
and by the way, this question that the pr fixed, it's a act that you designed? or it really a bug?

Pringled:
@docevilOck we ask any contributors to first discuss the issue with us, except for trivial changes, which this is not. Opening an issue and then immediately making a PR is not a discussion.

On the actual issue: the design here is intentional. The MCP server is designed to be multi-repo, which is why the tool instructions and README  both tell callers to pass repo explicitly. Defaulting to the MCP process cwd is risky: the server is spawned by the agent harness from whatever directory it happens to be in (which is often ~), so we'd silently pre-index and file-watch the wrong tree.

docevilOck:
> @docevilOck we ask any contributors to first discuss the issue with us, except for trivial changes, which this is not. Opening an issue and then immediately making a PR is not a discussion.
> 
> On the actual issue: the design here is intentional. The MCP server is designed to be multi-repo, which is why the tool instructions and README both tell callers to pass repo explicitly. Defaulting to the MCP process cwd is risky: the server is spawned by the agent harness from whatever directory it happens to be in (which is often ~), so we'd silently pre-index and file-watch the wrong tree.

ok, thanks for your reply, i got your mean now. i doesn't notice the rule in this doc before, sry about that.

PR #153, opened 2026-05-24
Fix sort key in hybrid search candidate ordering

hobostay:
## Summary
- Fixes the sort key in `search()` to sort by `(file_path, start_line)` instead of just `start_line`, matching the comment's stated intent
- Fixes typo "introduces" → "introduced" in the comment

## Problem
The comment on line 108-109 says "Sort by the file path and start line to counteract randomness introduced by hashing", but the actual sort key was only `c.start_line`. This means chunks from different files with the same start line number would have non-deterministic ordering.

## Test plan
- [x] Existing tests pass
- [x] The change is a one-character fix to a sort key — verified that `(c.file_path, c.start_line)` produces deterministic ordering across files

### DIFF
--- a/src/semble/search.py
+++ b/src/semble/search.py
@@ -106,10 +106,10 @@ def search(
     normalized_bm25 = _rrf_scores(bm25_scores)
 
     # Sort by the file path and start line to
-    # counteract randomness introduces by hashing.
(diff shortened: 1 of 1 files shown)


stephantul:

PR #155, opened 2026-05-24
Log errors when file watcher stops unexpectedly

hobostay:
## Summary
- Changed the outer except Exception: pass in _watch_loop to log a warning with traceback instead of silently swallowing errors

## Problem
The _watch_loop method had a bare except Exception: pass wrapping watchfiles.awatch(). If the watcher failed for any reason (deleted directory, permission errors, OS resource limits), it would silently stop with no indication. This made debugging production issues very difficult.

## Test plan
- [x] Existing tests pass
- [x] The change only replaces pass with a logger.warning() call


stephantul:

PR #156, opened 2026-05-24
Add --content support to index subcommand

hobostay:
## Summary
- Replaces the deprecated --include-text-files flag on the index subcommand with the modern --content argument
- Updated _run_index to accept content types directly
- Added test for --content with the index subcommand

## Problem
The index subcommand only supported the deprecated --include-text-files flag, while search and find-related subcommands already had the modern --content argument.

## Test plan
- [x] Updated existing test_run_index to use new signature
- [x] Added test_index_via_cli_with_content for --content support
- [x] Existing tests for index CLI still pass

🤖 Generated with [Claude Code

### DIFF
--- a/src/semble/cli.py
+++ b/src/semble/cli.py
@@ -83,12 +83,12 @@ def _mcp_main() -> None:
     asyncio.run(serve(args.path, ref=args.ref, content=content))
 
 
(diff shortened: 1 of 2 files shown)


stephantul:

PR #175, opened 2026-06-02
feat: auto-detect git repo from cwd when no repo specified

### LINKED ISSUE #173: Auto-detect git repo from cwd when no repo specified
## Problem

When running the MCP server inside a git checkout without passing an explicit `repo` argument or configuring a default index, it raises a `ValueError`. Users have to always specify the repo path explicitly, which is unnecessary friction for the common case of "I'm already in the repo."

## Proposed change

When neither `repo` nor `default_source` is set, walk up from `cwd` to find the nearest `.git` directory and use that as the source. This makes the MCP server work out of the box when run inside a git checkout.

Add a `find_git_root()` utility in `src/semble/utils.py` and call it as a fallback in `_get_index()` in `src/semble/mcp.py`.

## Files touched

- `src/semble/utils.py` — new `find_git_root()` function
- `src/semble/mcp.py` — fallback logic in `_get_index()`
- `tests/test_mcp.py` — test coverage

pikapi12321:
## Linked issue

Closes #173

## Summary

- Add `find_git_root()` utility in `src/semble/utils.py` — walks up from cwd to find nearest `.git`
- In `_get_index()`, fall back to detected git root when neither `repo` nor `default_source` is set
- MCP server now works without explicit repo config when run inside a git checkout

## Checklist

- [x] This PR is linked to an existing issue
- [x] `make test` passes
- [x] `make lint` and `make typecheck` pass
- [x] Added tests for behaviour changes

### DIFF
--- a/src/semble/mcp.py
+++ b/src/semble/mcp.py
@@ -16,7 +16,7 @@ from semble.cache import save_index_to_cache
 from semble.index import SembleIndex
 from semble.index.dense import load_model
 from semble.types import ContentType
-from semble.utils import format_results, is_git_url, resolve_chunk
+from semble.utils import find_git_root, format_results, is_git_url, resolve_chunk
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


stephantul:
Hey, thanks for the PR but this is not a feature we're interested in right now.

PR #176, opened 2026-06-02
feat: Add semble install

Pringled:
This PR adds a `semble install` command. It detects installed coding agents and configures semble globally across three integration types (MCP, instructions, sub-agent). We use tree-sitter for json/jsonc parsing so comments in MCP/settings jsons are preserved.

Here's what the flow looks like in the terminal on running`semble install` (`semble uninstall` gives the same flow but the reverse (uninstalling whatever is selected)):

Screenshot 2026-06-03 at 07 45 16
Screenshot 2026-06-03 at 07 45 20
Screenshot 2026-06-03 at 07 44 40
Screenshot 2026-06-03 at 07 45 03


Pringled:
@greptileai review

Pringled:
@greptileai review this branch again main. Focus on two things:
- e2e testing, ensuring that all the codepaths and combinations resolve correctly or have a clear error message when they fail
- architecture, ensuring that the architecture for the installer is maintainable, easy to understand, and not over-engineered

docs/installation.md
```diff
@@ -0,0 +1,257 @@
...
+- **[MCP server](#mcp-server)**: exposes Semble as a native tool your agent can call directly.
+- **[AGENTS.md](#instructions-agentsmd--claudemd)**: adds a Semble usage guide to the agent's config file (`CLAUDE.md`, `AGENTS.md`, etc.).
+- **[Sub-agent](#sub-agent)**: installs a dedicated `semble-search` sub-agent for harnesses that support it.
+
+Install the CLI with [uv](https://docs.astral.sh/uv/getting-started/installation/), then run:
+
+```bash
+uv tool install semble
+semble install           # configure your agents
+semble uninstall         # undo
```
stephantul: This is a bit weird, because the instruction seems to imply that you need to uninstall it immediately. Maybe make it a separate block? 

i.e., if I copy paste this block, I'll just install and uninstall.

src/semble/cli.py
```diff
@@ -52,7 +52,9 @@ def _maybe_save_index(index: SembleIndex, path: str) -> None:
 def _agent_path(agent: Agent) -> Path:
     """Return the project-relative path where the semble sub-agent file should be written."""
     base_dir = ".github" if agent is Agent.COPILOT else f".{agent.value}"
-    return Path(base_dir) / "agents" / "semble-search.md"
+    # Copilot requires the .agent.md extension; the others use a plain .md file.
```
stephantul: do the others not accept `.agent.md`? As in: copilot only reads `.agent.md` files, but maybe the others do as well. If so, could be useful to just keep it like this. But maybe a bit brittle

pyproject.toml
```diff
@@ -33,7 +33,8 @@ dependencies = [
     "pathspec>=0.12",
     "tree-sitter>=0.25,<0.26",
     "tree-sitter-language-pack>=1.0,<1.8.0,!=1.6.3",
-    "orjson"
+    "orjson",
+    "questionary>=2.0",
```
stephantul: maybe an upper bound on the pin as well? (i.e., <3.0)

src/semble/installer.py
```diff
@@ -0,0 +1,659 @@
...
+from dataclasses import dataclass
+from importlib.resources import files
+from pathlib import Path
+from typing import Callable, Literal, NoReturn, Sequence, TypeVar, cast
+
+import questionary
+from tree_sitter import Node, Parser
+from tree_sitter_language_pack import SupportedLanguage, download, get_parser
+
+_HOME = Path.home()
```
stephantul: Just out of caution: if this function does not work for some reason, this script will fail on import. I'm not sure if it ever will, just something to keep in mind.

To avoid this, you can create a function and potentially cache the result.

src/semble/installer.py
```diff
@@ -0,0 +1,659 @@
...
+]
+
+
+def _json5_parser() -> Parser | None:
+    """Return a tree-sitter JSON5 parser, downloading the grammar if needed.
+
+    "json5" ships in tree-sitter-language-pack but isn't in its typed language list, hence the cast.
+    """
+    try:
+        download(["json5"])
```
stephantul: we can potentially ship this grammar with the package. It's a useful exception. I don't know how large it is, but I suspect it's pretty small.

src/semble/installer.py
```diff
@@ -0,0 +1,659 @@
...
+        display_name="Zed",
+        binary="zed",
+        config_dir=_HOME / ".config" / "zed",
+        mcp=McpConfig(_HOME / ".config" / "zed" / "settings.json", "context_servers", _ZED_ENTRY),
+        instructions_path=None,
+    ),
+]
+
+
+def _json5_parser() -> Parser | None:
```
stephantul: I think this could be cached: if the download fails or times out, it will fail or time out every time.

src/semble/installer.py
```diff
@@ -0,0 +1,659 @@
...
+    "json5" ships in tree-sitter-language-pack but isn't in its typed language list, hence the cast.
+    """
+    try:
+        download(["json5"])
+        return get_parser(cast(SupportedLanguage, "json5"))
+    except Exception:
+        return None
+
+
+def _json5_object(text: str) -> JsonObjectResult:
```
stephantul: I'm not a big fan of these kinds of returns: they put a lot of work in the caller. Instead, you could consider returning a tuple containing 

(result, success)

if success is True, you can safely read result, if it is False, the result contains the reason for the error.

src/semble/installer.py
```diff
@@ -0,0 +1,659 @@
...
+5. Use Grep/Glob/Read only when you need exhaustive literal matches or quick confirmation of an exact string.
+{SEMBLE_END}
+"""
+
+
+@dataclass(frozen=True)
+class McpConfig:
+    """MCP integration config for one agent."""
+
+    path: Path | PathResolver
```
stephantul: As far as I can tell, there is no need for this disjunction. The paths can resolve when creating the `McpConfig` itself, e.g.

instead of doing:

```python
McpConfig(_vscode_mcp_path)
```

you can do:

```python
McpConfig(_vscode_mcp_path())
```

src/semble/installer.py
```diff
@@ -0,0 +1,659 @@
...
+_ACTION_DETAIL: dict[str, str] = {
+    "skipped": "JSON5 grammar unavailable — add manually (see README)",
+    "error": "could not parse or edit config",
+}
+
+# Config fragments
+SEMBLE_START = "<!-- SEMBLE_START -->"
+SEMBLE_END = "<!-- SEMBLE_END -->"
+
+_STDIO_ENTRY: dict[str, object] = {
```
stephantul: I think the use of `_ENTRY` is slightly ambiguous here. I first read it as "an entry" (i.e., a record), but I think you mean "entrypoint"? If so, maybe rename it to read entrypoint.

src/semble/installer.py
stephantul, line 1: One thing I think would really improve this code is moving the JSON5 and TOML parsing to a separate file or separate files. It would lead to a lot more readability. Likewise, the definitions can move to a separate `const` file? That way you separate the entry points for the TOML and JSON5 parsing from the rest of the code, and also separate the constants from operational code. I think just making a submodule called `installer` and adding those under this could be fine.

stephantul (approved):
Small docs comments, but looks good!

docs/installation.md
Pringled: Yeah agree, fixed

src/semble/cli.py
Pringled: Yanked semble init as discussed so this is also resolved

src/semble/installer.py
Pringled: Yup better safe than sorry, fixed

src/semble/installer.py
Pringled: I think this is a followup, it's a bit of work to do well

src/semble/installer.py
Pringled: Done

src/semble/installer.py
Pringled: The issue is that this can't resolve at definition time since VS Code and OpenCode paths are dynamic (OS-dependent), so I think we do need the callable right?

src/semble/installer.py
Pringled: Yeah agree, renamed it

src/semble/installer.py
Pringled: Yuuuuup I like this! Refactored it, lmk what you think

src/semble/installer.py
Pringled: Hmm I tried it but could not really come up with a nice way to do it, everything I tried ended up with #type ignores since the success flag does not narrow the result type. I tried a small additional dataclass but tbh that adds more complexity then it removes. But if you have a good way of doing it lmk

PR #180, opened 2026-06-04
chore: make functions public, remove callable paths

stephantul:
This PR makes some functions that were private public again (they were imported as private), and removes the union over Path and Pathresolver. Instead, we just resolve all paths. This simplifies the code.


Pringled (approved):
Ship it

PR #182, opened 2026-06-04
fix: disallow relative paths in user-supplied cache dir

stephantul:
This PR disallows relative paths as part of the user-supplied cache dir strategy. If the user supplies a  relative path, we revert to the default index, and show a warning.


Pringled (approved):
Good catch

PR #192, opened 2026-06-11
feat: add Command Code agent support

### LINKED ISSUE #189: Add Command Code agent detection and integration
## Problem

`semble install` does not detect Command Code (`cmd`), requiring users to manually configure the MCP server. Command Code fully supports MCP servers via `cmd mcp add` and stores config in `~/.commandcode/mcp.json` under the `mcpServers` key — the same stdio format used by most other supported agents.

## Why semble

Command Code is a terminal-based AI coding agent with a growing user base. Auto-detection would let users run `semble install` and select Command Code alongside their other agents, consistent with how Claude Code, Codex, Cursor, Gemini CLI, Kiro, OpenCode, GitHub Copilot, VS Code, Windsurf, Zed, Reasonix, and Pi are handled today.

For Command Code specifically, the benefit is significant — it spins up an explore agent that reads the majority of files in a repo for each task, consuming thousands of tokens. Semble's semantic search would let it find relevant code in ~2% of the tokens instead.

## Manual setup (current workaround)

Users can currently add semble manually:

cmd mcp add semble --scope user -- uvx --from "semble[mcp]" semble

Or with a direct binary:

cmd mcp add semble --scope user -- /path/to/semble

## Proposed solution

Add Command Code as a detected agent in `semble install`, following the same pattern as existing agents (Claude Code, Cursor, Codex, etc.).

**I'm happy to implement this.**

hubaibmahmood:
- Add commandcode.md sub-agent definition in src/semble/agents/
- Add AgentTarget entry in src/semble/installer/agents.py with MCP config (~/.commandcode/mcp.json), instructions (~/.commandcode/AGENTS.md), and sub-agent (~/.commandcode/agents/semble-search.md)
- Update docs/installation.md with Command Code MCP, sub-agent, and supported agents sections
- Add commandcode to parametrized merge_mcp test

## Linked issue

Closes [#189](https://github.com/MinishLab/semble/issues/189)

## Summary

Adds Command Code (`cmd`) as a supported agent target in the Semble installer, following the same pattern as PR [#183](https://github.com/MinishLab/semble/pull/183) (Reasonix/Pi support). Command Code stores MCP config in `~/.commandcode/mcp.json` under the `mcpServers` key, which matches the bare stdio format used by other agents like Copilot, Windsurf, and Pi.

## Checklist

- [x] I have read [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md)
- [x] This PR is linked to an existing issue (above)
- [x] `make test` passes
- [x] `make lint` and `make typecheck` pass (or `make pre-commit`)
- [x] Added or updated tests for any behaviour changes
- [x] Updated docstrings / docs for any public API changes


src/semble/installer/agents.py
```diff
@@ -236,7 +236,16 @@ def _vscode_mcp_path() -> Path:
...
+    AgentTarget(
+        id="commandcode",
+        display_name="Command Code",
+        binary=None,
+        config_dir=_HOME / ".commandcode",
+        mcp=McpConfig(_HOME / ".commandcode" / "mcp.json", "mcpServers", _BARE_STDIO_SERVER_CONFIG),
+        instructions_path=_HOME / ".commandcode" / "AGENTS.md",
+        subagent_path=_HOME / ".commandcode" / "agents" / "semble-search.md",
+    ),
+    ]
```
Pringled: Small indentation issue, this should be indented like https://github.com/MinishLab/semble/blob/main/src/semble/installer/agents.py#L239, I think `make lint` should pick this up?

src/semble/agents/commandcode.md
```diff
@@ -0,0 +1,42 @@
+---
+name: semble-search
+description: Code search agent for exploring any codebase. Use for finding code by intent, locating implementations, understanding how something works, or discovering related code. Prefer over bash/grep for any semantic or exploratory question.
+runAs: subagent
```
Pringled: I can't find this in the [command code docs](https://commandcode.ai/docs/core-concepts/custom-agents), is this actually something that's supported?

src/semble/agents/commandcode.md
```diff
@@ -0,0 +1,42 @@
...
+allowed-tools: bash, read_file
```
Pringled: Should be `tools` from what I can tell from the [command code docs](https://commandcode.ai/docs/core-concepts/custom-agents).

src/semble/agents/commandcode.md
```diff
@@ -0,0 +1,42 @@
...
+
+If `semble` is not on `$PATH`, use `uvx --from "semble[mcp]" semble` in its place.
+
+### Workflow
+
+1. Start with `semble search` to find relevant chunks. The index is built and cached automatically.
+2. Use `--content docs` for documentation, `--content config` for config files, or `--content all` for everything.
+3. Inspect full files only when the returned chunk does not give enough context.
+4. Optionally use `semble find-related` with a promising result's `file_path` and `line` to discover related implementations.
+5. Use bash/grep only when you need exhaustive literal matches or quick confirmation of an exact string.
```
Pringled: This is not consistent with the other agent instructions where we use "Use grep only when you need exhaustive literal matches or quick confirmation of an exact string.". Is there a specific reason for that?

Pringled (changes requested):
Mostly good to go, few issues with the subagent instruction

src/semble/installer/agents.py
hubaibmahmood: Good catch — you're right, `make lint` runs `ruff check` but doesn't include `ruff format --check`. Running `ruff format` separately catches this. Will fix in the next commit.

src/semble/agents/commandcode.md
hubaibmahmood: You're right, `runAs` is not in the Command Code docs — I'll remove it. The default behavior already runs custom agents as subagents, so the field is unnecessary. Confused it with another CLI my bad.

src/semble/agents/commandcode.md
hubaibmahmood: Agreed — the Command Code docs show tools as the correct field (e.g. tools: "glob, grep, read_file, think"), not allowed-tools. I'll fix it to tools: bash, read_file to match the documented format.

src/semble/agents/commandcode.md
hubaibmahmood: No specific reason — just carried over from the Reasonix template. I'll update it to match the existing convention: "Use grep only when you need exhaustive literal matches or quick confirmation of an exact string."

Pringled (approved):
LGTM, thanks for adding this!

PR #194, opened 2026-06-11
feat: index extensionless scripts by shebang

presempathy-awb:
## Problem

Semble decides what to index purely from the file **extension** — `walk_files`/`_walk` gate on `item.suffix`, and `detect_language` keys on `Path.suffix`. Extensionless executables are therefore never indexed, even though their content is ordinary code: a `bin/` CLI starting with `#!/usr/bin/env python3`, a `#!/usr/bin/env bash` helper, etc.

This is common in real repos — a tree I was searching had ~200 extensionless scripts under `bin/` (Python and bash), none of which semble could surface. Existing escape hatches don't cover it:

- `--content all` only widens the docs/config **extension** sets; it doesn't relax the extension requirement.
- A `.sembleignore` negation can't whitelist them either — the bypass in `_is_ignored` requires the pattern to have a suffix.

## Change

An opt-in shebang fallback, off by default:

- **`detect_language_from_shebang(text)`** (`index/files.py`) resolves the interpreter from a script's first line. It takes the *last* shebang token whose basename maps to a known interpreter, so it handles plain interpreters (`#!/bin/bash`), `env` indirection (`#!/usr/bin/env python3`), `env -S` launcher wrappers (`#!/usr/bin/env -S uv run --quiet python3` → `python`), version suffixes (`python3.12` → `python`), and flags.
- **`get_shebang_languages(types)`** returns the shebang-reachable languages within the requested content types, so shebang inclusion honours the same `--content` selection as extension indexing (`--content docs` → no scripts).
- **`walk_files()` / `_walk()`** gain an opt-in `shebang_languages` parameter; extensionless files whose shebang maps to one of those languages are also yielded. **The default is empty, so existing behaviour is unchanged.**
- **`create.py`** passes the shebang languages and uses `detect_language(path) or detect_language_from_shebang(source)`, so the file is chunked with the right grammar (line-chunking fallback otherwise, as today).
- **`cache.py`** passes the same `shebang_languages` in its cache-validation walk, so `current_files` matches the build walk and the incremental cache doesn't invalidate on every run.

Only extensionless files are probed (a bounded 256-byte first-line read), so there's no extra I/O for files that already carry a non-indexed extension. Symlinks remain skipped.

## Tests

- `tests/test_files.py`: `detect_language_from_shebang` across `env` / `env -S` / version-suffix / flag / non-shebang cases, plus `get_shebang_languages` content-type gating.
- `tests/test_file_walker.py`: extensionless scripts are included when their language is requested, excluded when it isn't, and the unchanged default (no `shebang_languages` → extensionless skipped).

Full suite: `280 passed`.

## Before / after

In a repo with extensionless `bin/` scripts, a natural-language query whose best match lives in such a script previously returned only unrelated *extensioned* files; with this change the script itself is indexed and ranks as expected, and sibling `bin/` scripts become searchable too.

### DIFF
--- a/src/semble/cache.py
+++ b/src/semble/cache.py
@@ -9,7 +9,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from semble.index.file_walker import walk_files
-from semble.index.files import FileStatus, get_extensions, get_file_status
+from semble.index.files import FileStatus, get_extensions, get_file_status, get_shebang_languages
 from semble.index.types import PersistencePath
 from semble.types import ContentType
 from semble.utils import is_git_url, resolve_model_name
@@ -124,11 +124,12 @@ def get_validated_cache(path: str, model_path: str | None, content: Sequence[Con
 
     write_time = metadata["time"]
     extensions = get_extensions(content)
+    shebang_languages = get_shebang_languages(content)
 
     path_as_path = Path(path).resolve()
     stored_files: list[str] = metadata.get("file_paths", [])
     current_files = []
-    for file_path in walk_files(path_as_path, extensions=extensions):
+    for file_path in walk_files(path_as_path, extensions=extensions, shebang_languages=shebang_languages):
         file_status = get_file_status(file_path, write_time)
         if file_status == FileStatus.NEWER:
             return None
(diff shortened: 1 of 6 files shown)


stephantul:
This PR does not adhere to our [contributing guidelines](https://github.com/MinishLab/semble/blob/main/CONTRIBUTING.md) and will be closed. You also do not use our template. Please open an issue, describe your problem first, and then we'll see whether you need to open a PR. Closing.

PR #195, opened 2026-06-12
fix: Prevent concurrent write corruption in savings.jsonl

Pringled:
Multiple MCP server processes appending to savings.jsonl simultaneously caused a small amount of lines to contain interleaved JSON. Added a lock for unix,  Windows can technically still write malformed lines but it won't break the CLI, just show a warning (if we want to do it for windows we need to add a dependency which I don't really want to do for this).

### DIFF
--- a/src/semble/stats.py
+++ b/src/semble/stats.py
@@ -59,8 +59,15 @@ def save_search_stats(
         stats_file = _get_stats_file()
         stats_file.parent.mkdir(parents=True, exist_ok=True)
         with stats_file.open("a") as f:
+            try:
+                import fcntl
+            except ImportError:
+                pass  # Windows has no fcntl, proceed without lock
+            else:
+                fcntl.flock(f, fcntl.LOCK_EX)
             f.write(json.dumps(record) + "\n")
     except OSError:
+        # If we can't write to the stats file, just skip the stats for this call
         pass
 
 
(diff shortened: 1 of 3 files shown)


src/semble/stats.py
```diff
@@ -59,6 +59,14 @@ def save_search_stats(
         stats_file = _get_stats_file()
         stats_file.parent.mkdir(parents=True, exist_ok=True)
         with stats_file.open("a") as f:
+            try:
+                import fcntl
```
stephantul: One issue I can see here is that this import is always attempted (and always fails) on Windows and systems that do not support it. 

Maybe move it somewhere else or cache the result of the import operation? I honestly don't know how long a failed import takes, but this is tried on every save.

src/semble/stats.py
```diff
@@ -59,6 +59,14 @@ def save_search_stats(
...
+
+                fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            except ImportError:  # pragma: no cover
+                pass  # Windows has no fcntl, write unlocked
+            except OSError:  # pragma: no cover
```
stephantul, line 80: Is there some way to distinguish between these failure cases? I.e., just not being supported or a lock contention? The former should not be retried.

stephantul (approved):
Some comments, not sure how easy it is.

stephantul (approved):

PR #198, opened 2026-06-16
feat: Optimize MCP and AGENTS.md instructions, add snippet_lines

Pringled:
This PR adds several changes and optimizations:
  - chunk_size halved (1500 to 750). Retrieval quality essentially flat on our benchmark but big token savings
  - snippet_lines=10 default on MCP tools so agents get the function signature + a few lines to confirm the location, rather than full 30–60 line chunks which saves even more tokens
  - Explicit self-correction guidance in the MCP tool description: call again with snippet_lines=None if the snippet isn't enough
  - Navigation instructions updated across all agent files so they go directly to the returned file and line and don't re-grep
  - Cache invalidation on chunk_size so stale index now rebuilds instead of silently returning wrong results

I tested these changes on SWE-bench Lite across 5 models (gpt-5.4, gpt-5.4-mini, deepseek-v4-pro, claude-sonnet-4.6, claude-haiku-4.5), and 3 harnesses (Claude Code, Codex, OpenCode). The biggest gains are for the MCP integration which consistently gave 50 to 70% cost reduction with flat or improved gold hit rate. For the CLI integrations (AGENTS.md/sub-agent) the savings were more modest (5 to 15% savings).


src/semble/agents/gemini.md
```diff
@@ -38,6 +38,6 @@ If `semble` is not on `$PATH`, use `uvx --from "semble[mcp]" semble` in its plac
 
 1. Start with `semble search` to find relevant chunks. The index is built and cached automatically.
 2. Use `--content docs` for documentation, `--content config` for config files, or `--content all` for everything.
-3. Inspect full files only when the returned chunk does not give enough context.
+3. Navigate directly to the returned file and line — do not re-search or grep for the same content.
```
stephantul: 🛑  em-dash detected 🛑

src/semble/agents/kiro.md
```diff
@@ -9,9 +9,9 @@ tools:
 Use `semble search` to find code by describing what it does or naming a symbol/identifier, instead of grep:
 
 ```bash
-semble search "authentication flow" ./my-project
-semble search "save_pretrained" ./my-project
-semble search "save model to disk" ./my-project --top-k 10
+semble search "authentication flow" ./my-project --snippet-lines 10  # signatures only, fast
```
stephantul: perhaps make this more literal? e.g.,

```
# first 10 lines only, fast
```

Also "fast" is debatable, it's not necessarily faster. Maybe "concise" actually fits better?

src/semble/cli.py
```diff
@@ -175,6 +178,13 @@ def _cli_main() -> None:
     search_p.add_argument("query", help="Natural language or code query.")
     search_p.add_argument("path", nargs="?", default=".", help="Local path or git URL (default: current directory).")
     search_p.add_argument("-k", "--top-k", type=int, default=5, help="Number of results (default: 5).")
+    search_p.add_argument(
```
stephantul, line 181: This argument, `snippet_lines`, is actually not that great of a description. Without knowing what it does, it's actually kind of surprising that:

1. this is a numeric argument
2. it truncates a chunk to a specific line length

So something like `n_snippet_lines` is maybe clearer. Or `lines_per_chunk`. 

Nevertheless I'm ok with keeping it, I know you did all the tests with this argument probably.

stephantul (approved):
Nice!

First off. The em dash comment is not real (it can't hurt you) (I just don't like em dashes)

I think both other comments might require rerunning experiments, so feel free to ignore them, and mostly are about how a human might read or use semble, not an agent.

PR #206, opened 2026-06-19
fix: Ensure savings are correct when max_snippet_lines is used

Pringled:
When `max_snippet_lines` was used, savings were undercounted. This PR fixes that.


stephantul (approved):
OK BY ME

PR #208, opened 2026-06-21
feat: make repo mandatory, remove default source & ref

stephantul:
This PR makes the `repo` argument mandatory for the MCP. The MCP used to take a `path` and `ref` argument on startup, but since MCP startup is hardcoded (not controlled by the agent/harness), it was difficult to use this parameter. The consequence of this was that it was possible for agents to send queries without a "repo", as the schema allowed this. This then led to exceptions in the MCP server (i.e., the calls returned no data, just an error message).

The one use case for adding a path, which is enterprise hosting of MCP servers, remains, but should be turned into another MCP server, which we'll add later. 

Of note: no docs changes were needed, since this `path` option was not even mentioned there.


tests/test_cli.py
```diff
@@ -14,7 +14,6 @@
 @pytest.mark.parametrize(
     "argv",
     [
-        ["semble", "/some/path", "--ref", "main"],
```
stephantul, line 17: I left this as a parameterize, even though there's only a single option now , seemed sensible given that we want to expand this again.

src/semble/types.py
```diff
@@ -60,13 +60,6 @@ class SearchResult:
     chunk: Chunk
     score: float
 
-    def to_dict(self) -> dict[str, Any]:
```
stephantul, line 63: This was somehow a pre-existing piece of dead code. Strange!

stephantul:
@greptileai review

src/semble/mcp.py
```diff
@@ -218,22 +199,6 @@ def _build_and_cache_index(self, source: str, ref: str | None, model_path: str,
     def evict(self, source: str) -> None:
         self._tasks.pop(self._compute_cache_key(source), None)
 
-    async def start_watcher(self, path: str) -> None:
```
Pringled, line 221: Maybe I'm missing something, but by removing this the MCP index never updates anymore I think?

Pringled (changes requested):
One concern

PR #211, opened 2026-06-23
fix: Validate cached MCP indexes on query

Pringled:
This PR ensures MCP indexes get validated on every query like the CLI route. There's a minimum time (3x original indexing time) to ensure we don't start re-indexing while it's still indexing).


Pringled:
@greptileai review

stephantul (approved):

PR #221, opened 2026-07-08
fix: pin generated MCP uvx package spec (fix #186)

### LINKED ISSUE #186: [suggestion] semble install should set exact versions for uvx commands
Currently, semble install will set the mcp as `uvx --from 'semble[mcp]' semble`, and will suggest `uvx` fallbacks in AGENTS.md and subagents.

This, however, leaves users vulnerable to supply chain attacks. If someone takes over the package, agents will install and execute the compromised package.

Semble install should pin to its own version.

koriyoshi2041:
## Problem

Generated MCP configs and agent instructions currently use `uvx --from "semble[mcp]" semble`, which resolves the latest published package each time uv refreshes its cache. That can make installed agent configs drift away from the Semble version that generated them, and it leaves the setup exposed to unexpected or compromised future releases.

Closes #186.

## Fix

- Build MCP config entries from the current Semble package version, e.g. `semble[mcp]@0.4.2`.
- Reuse the same package spec for the Codex TOML installer block.
- Update manual install docs and packaged agent instruction templates to show pinned `uvx` commands.
- Add an installer regression test that checks every generated MCP config pins the current package version.

## Tests

- `uv run --extra dev pytest tests/test_installer.py`
- `uv run --extra dev ruff check src/semble/installer/agents.py src/semble/installer/config.py tests/test_installer.py`
- `uv run --extra dev mypy src/semble/installer/agents.py src/semble/installer/config.py`
- `git diff --check`

## Risk

Low. This only changes generated/manual `uvx --from` package specs; users can still upgrade Semble deliberately with `uv tool upgrade semble` and then reinstall/regenerate configs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ semble search "deployment guide" ./my-project --content docs   # or: config, all
 semble find-related src/auth.py 42 ./my-project
 ```
 
-`--content` accepts `code` (default), `docs`, `config`, or `all`. `path` defaults to the current directory when omitted; git URLs are accepted. If `semble` is not on `$PATH`, use `uvx --from "semble[mcp]" semble` in its place.
+`--content` accepts `code` (default), `docs`, `config`, or `all`. `path` defaults to the current directory when omitted; git URLs are accepted. If `semble` is not on `$PATH`, use `uvx --from "semble[mcp]@0.4.2" semble` in its place.
 
 <details>
 <summary>Controlling which files are indexed</summary>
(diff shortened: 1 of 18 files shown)


stephantul:
Hey @koriyoshi2041,

Thanks for this PR. This is not the solution we're looking for currently, because it requires us to update this every time we bump the version. For reference, in the future, should you want to solve up an issue, please let us know what you intend to do, so we can tell you what our own plans are, and we can tell you whether we are working on a parallel solution

koriyoshi2041:
Understood. I should have checked the intended release/versioning path before implementing this. I’ll confirm the approach on the issue first for future changes here.

PR #225, opened 2026-07-16
feat: Add partial reindexing

Pringled:
This PR adds partial reindexing support. BM25s is dropped in favor of a custom BM25 implementation that supports adding/removing documents without full rebuilds. Results on our benchmarks:

| Metric | Before | After | Change |
|---|---:|---:|---:|
| NDCG@10 | 0.8519 | 0.8519 | No change |
| Cold index build | 836 ms | 809 ms | 3.2% faster |
| Warm query p50 | 1.69 ms | 2.33 ms | 38% slower |
| Warm query p90 | 6.53 ms | 7.41 ms | 13% slower |

The scores are basically the same. Queries are a bit slower but still so fast that I don't think it matters. Now the big improvement, tested on Zig (one of the larger repos in our benchmark):

| Operation | Before | After | Change |
|---|---:|---:|---:|
| Rebuild after one changed file | 8.47 s | 369 ms | ~23x faster |


src/semble/index/create.py
```diff
@@ -2,57 +2,129 @@
...
-        args = BasicArgs()
-        semantic_index = SelectableBasicBackend(embeddings, args)
     else:
-        raise ValueError(f"No supported files found under {path}.")
+        changed_embeddings = embed_chunks(model, changed_chunks)
+        offset = 0
+        for index, _, count in embedding_parts:
+            vector_parts[index] = changed_embeddings[offset : offset + count]
+            offset += count
+        same_vector_layout = len(manifest) == len(previous_manifest) and all(
```
stephantul: Maybe turn this into a function

src/semble/index/index.py
```diff
@@ -310,33 +318,45 @@ def load_from_disk(cls: type[SembleIndex], path: Path | str) -> SembleIndex:
...
             raise FileNotFoundError(f"Index not found at {path}. Missing: {missing}")
 
-        bm_25_index = BM25.load(persistence_paths.bm25_index)
-        semantic_index = SelectableBasicBackend.load(persistence_paths.semantic_index)
         with open(persistence_paths.metadata, "rb") as f:
             metadata = orjson.loads(f.read())
+        found_version = metadata.get("cache_version")
+        if found_version != CACHE_FORMAT_VERSION:
+            raise ValueError(
+                f"Unsupported index format {found_version!r}; expected {CACHE_FORMAT_VERSION}. Rebuild the index."
```
stephantul: Maybe make it clear to the user how to rebuild the index? Or do it for them automatically? A possible risk with raising here is that an agent runs into this error and keeps searching and never actually runs semble anymore.

src/semble/index/index.py
```diff
@@ -310,33 +318,45 @@ def load_from_disk(cls: type[SembleIndex], path: Path | str) -> SembleIndex:
...
+
+        bm25_index = BM25.load(persistence_paths.bm25_index)
+        semantic_index = SelectableBasicBackend.load(persistence_paths.semantic_index)
         with open(persistence_paths.chunks, "rb") as f:
             chunk_data = orjson.loads(f.read())
 
         chunks = []
         for chunk_item in chunk_data:
             chunks.append(Chunk.from_dict(chunk_item))
+        if len(chunks) != len(bm25_index.doc_order) or len(chunks) != semantic_index.vectors.shape[0]:
```
stephantul: you can chain this: if not (len(chunks) == x == y)

src/semble/cache.py
```diff
@@ -145,3 +152,58 @@ def get_validated_cache(path: str, model_path: str | None, content: Sequence[Con
...
+            return None
+        manifest = {indexed_path: FileManifestEntry(**entry) for indexed_path, entry in raw_manifest.items()}
+
+        with open(persistence_path.chunks, "rb") as f:
+            chunks = [Chunk.from_dict(item) for item in orjson.loads(f.read())]
+
+        vectors = SelectableBasicBackend.load(persistence_path.semantic_index).vectors
+        bm25_index = BM25.load(persistence_path.bm25_index)
+        chunk_count = len(chunks)
+        if vectors.shape[0] != chunk_count or len(bm25_index.doc_order) != chunk_count:
```
stephantul: this is a repetition of a snippet I also commented on, maybe turn it into a function

src/semble/cache.py
```diff
@@ -145,3 +152,58 @@ def get_validated_cache(path: str, model_path: str | None, content: Sequence[Con
...
+                or any(chunk.file_path != indexed_path for chunk in chunks[entry.start : entry.start + entry.count])
+            ):
+                return None
+            expected_ids.extend(make_chunk_id(indexed_path, slot) for slot in range(entry.count))
+            next_start += entry.count
+        if next_start != chunk_count or bm25_index.doc_order != expected_ids:
+            return None
+
+        return PreviousIndex(chunks=chunks, vectors=vectors, manifest=manifest, bm25_index=bm25_index)
+    except (OSError, orjson.JSONDecodeError, json.JSONDecodeError, KeyError, TypeError, ValueError):
```
stephantul, line 206: maybe turn this into a constant? I'm not sure if that's the right way. Also maybe we expect this to log something useful? This seems to point towards an unexpected path, rather than an expected cache invalidation.

src/semble/index/types.py
```diff
@@ -28,3 +38,22 @@ def from_path(cls: type[PersistencePath], path: Path) -> PersistencePath:
...
         )
+
+
+@dataclass
+class FileManifestEntry:
+    """Record a file's modification time and chunk range within the global chunk list."""
+
+    mtime: float
+    start: int
+    count: int
```
stephantul, line 49: should this not be length? As in `start + length` == `end`?

src/semble/cache.py
```diff
@@ -145,3 +152,58 @@ def get_validated_cache(path: str, model_path: str | None, content: Sequence[Con
...
+        bm25_index = BM25.load(persistence_path.bm25_index)
+        chunk_count = len(chunks)
+        if vectors.shape[0] != chunk_count or len(bm25_index.doc_order) != chunk_count:
+            return None
+        expected_ids: list[str] = []
+        next_start = 0
+        for indexed_path, entry in manifest.items():
+            if (
+                entry.start != next_start
+                or entry.count < 0
```
stephantul: when can entry.count be < 0? Should this not fail when creating the manifest? It seems to me that a manifest item with a negative count can't be valid anyway.

src/semble/index/create.py
```diff
@@ -2,57 +2,129 @@
...
-            source = read_file_text(file_path)
-            chunk_path = file_path.relative_to(display_root) if display_root else file_path
-            chunks.extend(chunk_source(source, str(chunk_path), language))
 
-    if chunks:
+            indexed_path = str(file_path.relative_to(display_root) if display_root else file_path)
+            mtime = file_path.stat().st_mtime
+            previous_entry = previous_manifest.get(indexed_path)
+
+            if previous is not None and previous_entry is not None and previous_entry.mtime == mtime:
```
stephantul: The previous_entry.mtime is a float comparison. I wonder if this ever doesn't work?

src/semble/index/create.py
```diff
@@ -2,57 +2,129 @@
...
+                file_chunks = previous.chunks[previous_entry.start : previous_entry.start + previous_entry.count]
```
stephantul: maybe create an `end` property to be used. e.g., `end` is defined as `start + count`.

src/semble/index/create.py
```diff
@@ -2,57 +2,129 @@
...
+                vector_parts.append(
+                    previous.vectors[previous_entry.start : previous_entry.start + previous_entry.count]
+                )
+            else:
+                source = read_file_text(file_path)
+                file_chunks = chunk_source(source, indexed_path, language)
+                _reindex_file(bm25_index, indexed_path, file_chunks, previous_entry)
+
+                embedding_parts.append((len(vector_parts), len(chunks), len(file_chunks)))
+                vector_parts.append(np.empty((0, model.dim), dtype=np.float32))
```
stephantul: I don't understand this yet, why do we need an empty thing here?

stephantul (approved):
Looks good to me! Lots of small comments, but nothing major.

src/semble/index/create.py
Pringled: Agree, I turned this into a helper

src/semble/index/index.py
Pringled: Yeah makes sense, I've updated the error message

src/semble/cache.py
Pringled: I've added a debug log with the exception context before falling back, that should be helpful.I'm not sure about a constant though, this is only used once here and I think it's pretty specific for the cache

src/semble/index/types.py
Pringled: Hmm it represents the number of chunks so I think count is a bit more intuitive

src/semble/index/create.py
Pringled: To make batching work across different files (you need to preserve the fileorder), but tbh, this is kind of overtuned, I'll just embed per file. The chances of having to re-embed multiple files is small and even if you have to do it it's superfast

src/semble/cache.py
Pringled: Yeah you're right this is overly defensive, we control the cache so this won't ever happen in practice, I'll get rid of it

src/semble/index/create.py
Pringled: Yess good catch, will update

src/semble/index/create.py
Pringled: I like it, done

PR #227, opened 2026-07-22
feat: add --version flag

stephantul:
This PR adds a little version flag that can be used to check whether your semble is up to date.

```
semble --version
semble -V
```

It just prints `__version__`, so not a lot going on. In the future, we can enrich this using other metadata (builds etc.)


Pringled (approved):
LGTM

PR #230, opened 2026-07-22
Expose raw semantic cosine scores in search results

yasyf:
Hybrid search currently returns only the fused RRF score, which reflects rank position rather than absolute semantic relevance — a top-ranked result over weak matches is indistinguishable from a genuinely strong hit (RRF caps near 1/61 per leg before boosts).

This adds a `semantic_score` field to `SearchResult` carrying the dense leg's raw cosine similarity, threaded through fusion and both reranking paths without changing ranking behavior. BM25-only results expose `semantic_score: None`. The field is included in CLI and MCP output.

Motivation: agent consumers (we drive semble from the ccx/cc-context MCP facade) need an absolute relevance signal to detect all-weak result sets instead of mistaking top-ranked junk for evidence.

- Additive only; no ranking or API behavior changes
- Exact known-distance tests cover the cosine conversion (a sign inversion now fails), plus BM25-only `None` coverage
- `pytest` green, 100% coverage maintained, pre-commit hooks pass

### DIFF
--- a/src/semble/search.py
+++ b/src/semble/search.py
@@ -32,7 +32,10 @@ def _search_semantic(
     query_embedding = model.encode([query])
     indices, scores = semantic_index.query(query_embedding, k=top_k, selector=selector)[0]
     # Vicinity returns cosine distance; convert to similarity so higher = better.
-    return [SearchResult(chunk=chunks[index], score=1.0 - float(distance)) for index, distance in zip(indices, scores)]
+    return [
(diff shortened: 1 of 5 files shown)


stephantul:
Hey, please read the [CONTRIBUTING.md](https://github.com/MinishLab/semble/blob/main/CONTRIBUTING.md) and open an issue first. Also, we use PR templates for a reason, please don't ignore them. Closing for now.

Issue #224, opened 2026-07-16
Grammar download failure reported as "Language not found" and silently degrades chunking; support tree-sitter-* wheels


pszemraj:
Follow-on from #94. Catching the exception fixed the crash, but semble now silently falls back to line chunking and the caller never finds out.

Clean repro. Box behind a TLS-intercepting proxy with no route to github.com:

```
$ python -c "from tree_sitter_language_pack import download; download(['python'])"
DownloadError: Download error: Failed to fetch manifest from
https://github.com/kreuzberg-dev/tree-sitter-language-pack/releases/download/v1.6.2/parsers.json:
io: invalid peer certificate: UnknownIssuer

$ pip install tree-sitter-python tree-sitter-cpp tree-sitter-c
Successfully installed tree-sitter-c-0.24.2 tree-sitter-cpp-0.23.4 tree-sitter-python-0.25.0
```

Same box, same proxy, a minute apart. pip is allowed, the pack's downloader isn't. Every environment I run semble in looks like this: package mirror, no general egress. Staging the cache with `download()` on a machine that has egress and copying it over works, but that's a second machine, a build step, and an ABI nothing pins. There's no in-process path.

To reproduce: install semble, `rm -rf ~/.cache/tree-sitter-language-pack`, run a search from a host with no egress to github.com[^1]

[^1]: Literally how I ran into this was from Claud.Ai where you can whitelist package managers and so on, but random GitHub release URLs and other stuff does not work for downloads

```
$ semble search "complex exponential moving average" -k 5 . 2>&1 >/dev/null
Language python not found, falling back to line chunking
Language cpp not found, falling back to line chunking
Language c not found, falling back to line chunking
Language cuda not found, falling back to line chunking
$ echo $?
0
```

That message points the wrong way: `get_parser("python")` raises `LanguageNotFoundError`, so `chunking/core.py:37` logs "not found", but the message on the exception is `Download error: Failed to fetch manifest from ...`. Python isn't missing, the network is. Nothing in the output says so unless you print the exception yourself; logging `str(e)` there would cover it I think (with a path forward)

Cost is smaller than I claimed last time. the Ranking barely moves: six queries over a 43-file repo with and without parsers, same top-hit file 6/6, scores within 0.002  BUT What changes is where chunks get cut:

```
line chunking: 329 chunks,  30 start at a definition   (9.1%)
AST chunking : 378 chunks,  81 start at a definition  (21.4%)
```

[...]

PR #231, opened 2026-07-25
fix: preserve parser load failure details

koi-lee:
## Summary

- include the underlying `LanguageNotFoundError` or `DownloadError` detail in parser fallback warnings
- keep the existing fallback to line chunking unchanged
- add regression coverage for wrapped manifest-download and certificate failures

## Why

As described in #224, `tree-sitter-language-pack` can surface a download or TLS failure as `LanguageNotFoundError`. The current warning drops the exception detail, so users only see that a language was not found and cannot identify the blocked manifest download.

This is intentionally limited to the diagnostics portion of #224; it does not change grammar packaging or add dependencies.

## Validation

- `uv run pytest tests/test_chunker.py::test_get_parser tests/test_chunker.py::test_language_not_found_reports_download_failure tests/test_chunker.py::test_download_error -q`
- `uv run ruff check src/semble/chunking/core.py tests/test_chunker.py`
- `uv run mypy src/semble/chunking/core.py`

The full `tests/test_chunker.py` run was not completed because the uncached parser tests waited on the runtime grammar download. The affected failure paths are covered with mocks above.

### DIFF
--- a/src/semble/chunking/core.py
+++ b/src/semble/chunking/core.py
@@ -33,10 +33,18 @@ def _cached_get_parser(language: SupportedLanguage) -> Parser | None:
     """Gets a parser from tree_sitter."""
     try:
         return get_parser(language)
-    except LanguageNotFoundError:
-        logger.warning("Language %s not found, falling back to line chunking", language)
-    except DownloadError:
-        logger.warning("Failed to download language %s, falling back to line chunking", language)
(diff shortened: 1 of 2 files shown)


Pringled:
Hi @koi-lee, we are currently reworking this part of Semble and not accepting PR's for this. For the next time, please look at our contributing guidelines and use our template.

PR #238, opened 2026-08-01
feat: Switch to semble-grammars

Pringled:
This PR swaps out tree-sitter-language-pack for https://github.com/MinishLab/semble-grammars. This means the binaries become ~8mb bigger, but we get:
  - No runtime downloads since all grammars are bundled, so there's no network calls after installing (except the one time model download)
  - Compiled language packs for all platforms (including Intel macs and Arm Windows builds which we didn't have before)
  - Control over what is and isn't packaged 
  - Bit easier error handling
  - One less dependency

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ dependencies = [
     "tree-sitter>=0.25,<0.26",
     "orjson",
     "questionary>=2.0,<3.0",
-    "semble-grammars>=0.1.1",
+    "semble-grammars>=0.1.2",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -145,3 +146,6 @@ exclude_also = ["if __name__ == .__main__.:", "@abstractmethod"]
 
 [tool.uv]
 exclude-newer = "1 week"
+
+[tool.uv.exclude-newer-package]
+semble-grammars = "2026-08-02T00:00:00Z"
```
stephantul: You can replace this with 

```
exclude-newer-package = { semble-grammars = false }
```
Under tool.uv. 

Or just set this to `false` I guess.

This makes sense to me: we control semble-grammars, so I don't see the supply chain risk anyway.

stephantul (approved):
NIIIICE 🚀

pyproject.toml
Pringled: ohh nice, done!

Issue #246, opened 2026-08-10
MCP search does not expose the --content filter documented in generated AGENTS.md


markjfisher:
## Summary

Semble 0.5.4 generates `AGENTS.md` instructions telling agents to use `--content docs`, `--content config`, or `--content all` with `mcp__semble__search`.

However, the MCP tool schema exposes only:

- `query`
- `repo`
- `top_k`
- `max_snippet_lines`

There is no `content` argument.

Adding `--content docs` to the `query` string does not apply a filter. It is treated as ordinary query text, and the search still defaults to code.

## Expected behaviour

The MCP tool should expose a `content` parameter equivalent to the CLI option, for example:

```json
{
  "repo": "/path/to/repository",
  "query": "packet-native channel framing",
  "content": "docs"
}
```

Alternatively, the generated AGENTS.md should make clear that content filtering is available only through the CLI.

## Actual behaviour

A documentation query through MCP returns source-code results because the content type cannot be selected.

The equivalent CLI command works correctly:

```cli
semble search \
    "packet-native channels bypass SLIP while RS-232 retains stream framing" \
    . \
    --content docs
```

It returns the expected Markdown documentation.

## Environment

- Semble: 0.5.4
- Integration: MCP
- Platform: Linux
- Generated MCP server command:
  `uvx --from "semble[mcp]==0.5.4" semble`

## Additional observation

After restarting stale MCP server processes, search performance was approximately:

- cold MCP query: 12.5 seconds
- identical warm query: 3.7 seconds

Pringled:
Hi @markjfisher, thanks for creating this issue, this is a good catch, we'll add this in our next release. Note that you can use --content with MCP, but it's a startup instruction, not a query instruction, which is indeed nicer. Will let you know once it's there

Pringled:
Hey @markjfisher, this is now released with https://github.com/MinishLab/semble/pull/247. Let us know if you encounter any issues!

markjfisher:
Excellent. I'll pull this down and test it. Thanks!

PR #247, opened 2026-08-11
feat: add content filter to MCP

Pringled:
This PR adds a `content` argument to MCP, just like the CLI, resolving https://github.com/MinishLab/semble/issues/246. Searches still default to code. When docs, config, or all is explicitly requested, the MCP lazily builds an index containing exactly that content and caches it separately for reuse. This avoids making the default indexing path slower or unexpectedly indexing large non-code files.

### DIFF
--- a/src/semble/cache.py
+++ b/src/semble/cache.py
@@ -37,7 +37,7 @@ def cache_key(path: str) -> str:
 def find_index_from_cache_folder(path: str, content: Sequence[ContentType] = (ContentType.CODE,)) -> Path:
     """Find an exact content index in the cache for a project path."""
     cache_dir = resolve_cache_folder() / cache_key(path)
-    scope = "-".join(content_type.value for content_type in ContentType if content_type in content)
+    scope = "-".join(sorted({content_type.value for content_type in content}))
     return cache_dir / ("index" if scope == ContentType.CODE.value else f"index-{scope}")
 
 
--- a/src/semble/cache.py
+++ b/src/semble/cache.py
@@ -34,10 +34,11 @@ def cache_key(path: str) -> str:
     return hashlib.new("sha256", data).hexdigest()
 
 
-def find_index_from_cache_folder(path: str) -> Path:
-    """Finds an index from a cache folder and a project path."""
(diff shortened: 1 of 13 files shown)


src/semble/cache.py
```diff
@@ -34,10 +34,11 @@ def cache_key(path: str) -> str:
     return hashlib.new("sha256", data).hexdigest()
 
 
-def find_index_from_cache_folder(path: str) -> Path:
-    """Finds an index from a cache folder and a project path."""
+def find_index_from_cache_folder(path: str, content: Sequence[ContentType] = (ContentType.CODE,)) -> Path:
+    """Find an exact content index in the cache for a project path."""
     cache_dir = resolve_cache_folder() / cache_key(path)
-    return cache_dir / "index"
+    scope = "-".join(content_type.value for content_type in ContentType if content_type in content)
```
stephantul: Maybe a sort would be safer here. I can imagine us adding things  which might change the Enum order. (I think Enums are safely sortable though).

src/semble/mcp.py
```diff
@@ -27,22 +27,32 @@
 
 _CACHE_MAX_SIZE = 10  # Max number of cached indexes to keep in memory
 _MIN_REVALIDATE_FACTOR = 3  # Don't recheck staleness sooner than this many times the last build's duration
+ContentSelection = Literal["code", "docs", "config", "all"]
```
stephantul, line 30: Can this be derived from the Enum? Not sure if it makes sense to do so.

stephantul (approved):
Two small and speculative comments, but looks good to me otherwise regardless, thanks!

src/semble/cache.py
Pringled: Oh yeah that's a bit nicer, willdo

src/semble/mcp.py
Pringled: Yeah I thought about this but the answer is... not really, because the type would need to be a union of ContentType and the new "all", but we also cannot add "all" to contentype because it's not a contenttype.

PR #251, opened 2026-08-17
Make max file size configurable via SEMBLE_MAX_FILE_BYTES and warn on skipped files

### LINKED ISSUE #250: Files larger than 1 MB are silently skipped during indexing - no warning, no way to raise the limit
## Summary

Files larger than 1 MB (`_MAX_FILE_BYTES` in `src/semble/index/files.py`) are skipped during indexing **without any warning**, and there is no way to raise or lower the limit. The result is a silent coverage gap: queries against content that only exists in large files return plausible-looking results from *other* files, with no indication anything is missing.

## Version

semble 0.5.5 (also present on current `main`).

## Reproduction (synthetic)

```bash
mkdir demo && cd demo
python -c "
from pathlib import Path
body = '\n'.join(f'Paragraph {i} with unrelated filler text.' for i in range(30000))
Path('bigdoc.md').write_text('# UNIQUE.HEAD.42 widget alignment protocols\n\n' + body)
Path('small.md').write_text('# Small doc\n\nText about procedural compliance and unrelated filler.\n')
"
# bigdoc.md is ~2 MB; small.md is small
 semble search "widget alignment protocols" . --content all
```

**Observed:** only `small.md` is returned. Nothing on stderr or stdout mentions that `bigdoc.md` was skipped — the search looks successful, just wrong.

The existing test suite even documents this as intended behavior ("Files exceeding _MAX_FILE_BYTES are silently skipped during indexing"), so the skip itself is by design — the *silence* is the problem.

## Root cause

- `src/semble/index/files.py:8` — `_MAX_FILE_BYTES = 1_000_000`, a hard constant; no env var, config file, or CLI flag overrides it.
- `src/semble/index/create.py` — `create_index_from_path` skips any file whose status is not `VALID` with a bare `continue`; `TOO_LARGE` is never surfaced.
- `src/semble/cache.py` — cache validation skips the same files just as silently.

## Why it matters

For agents (the tool's primary audience), plausible-but-incomplete search results are worse than an error: there is no signal that trust in the results should be discounted. The failure mode is especially costly because large files are often generated or ingested documents — precisely the material an agent is asked to search.

## Suggested remedies (in preference order)

1. **Make the cap configurable** — e.g. a `SEMBLE_MAX_FILE_BYTES` env var and/or a CLI flag, following the existing `SEMBLE_CACHE_LOCATION` / `SEMBLE_CLONE_TIMEOUT` / `SEMBLE_MODEL_NAME` idiom.
2. **Warn at index time** — log a `WARNING` listing skipped `TOO_LARGE` files so the gap is discoverable.
3. Optionally, chunk-stream large files instead of skipping them whole.

Separately — and purely as a soft suggestion, not something this report depends on — it may be worth reconsidering whether 1 MB is the right default: embeddings are a model2vec `StaticModel` lookup (not transformer inference) and indexes are disk-cached with incremental rebuild, so a larger default cap looks cheap in practice. Happy to leave the default exactly where it is if the lean-build posture is preferred; configurability alone resolves the practical problem.

TommyC81:
Closes #250

Following the discussion in #250, this makes the 1 MB max file size configurable via a `SEMBLE_MAX_FILE_BYTES` environment variable and makes skipped files visible — no CLI flag, no default change.

## Changes

- **`SEMBLE_MAX_FILE_BYTES` env var** (`src/semble/index/files.py`) — new `get_max_file_bytes()` resolves the limit per call from the environment, falling back to the unchanged 1 MB default. Follows the existing `SEMBLE_CACHE_LOCATION` / `SEMBLE_CLONE_TIMEOUT` / `SEMBLE_MODEL_NAME` idiom, and works for the MCP server path too, where env vars are the only channel.
- **Input validation** — a malformed or nonpositive value warns and falls back to the default instead of crashing indexing.
- **Skip warning at index time** (`src/semble/index/create.py`) — `create_index_from_path` collects `TOO_LARGE` files and logs one `WARNING` naming them (up to 5, then `...`) with the active limit. The CLI installs an idempotent stderr handler on the `semble` logger so the warning is visible; the root logger is untouched.
- **README** — one paragraph under *Storage* documenting the 1 MB skip, the warning, and the override.
- **Tests** — four tests in `tests/index/test_index.py`: env resolution (default + override), invalid values fall back, oversized file skipped *with* a warning naming the file, and oversized file indexed once the limit is raised.

## Checklist

- [x] Links to an existing issue (`Closes #250`)
- [x] `make test` — 328 passed (4 new)
- [x] `make lint` / `make typecheck` — ruff check, ruff format, mypy clean
- [x] Behavior change covered by tests
- [x] No public API change; docstrings updated where touched
- [x] Focused diff (env var + warning, nothing else)

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,8 @@ Savings are calculated as follows: for each call, semble records the total chara
 
 By default, your Semble savings statistics and any saved indexes are stored in the OS cache folder (`~/Library/Caches/semble/` on macOS, `~/.cache/semble/` on Linux, `%LOCALAPPDATA%\semble\Cache\` on Windows). To override this location you can supply an environment variable `SEMBLE_CACHE_LOCATION` which should be the full path to the target cache location e.g. `~/my-folder/my-caches/semble`.
 
+Files larger than 1 MB are skipped during indexing to keep index builds lean. Skipped files are reported as a warning at index time. If you work with large generated or ingested documents, you can raise (or lower) this limit with the `SEMBLE_MAX_FILE_BYTES` environment variable (in bytes), or per invocation with `semble search --max-file-bytes <BYTES>`.
+
 On first use, Semble also downloads the embedding model from Hugging Face and caches it in the standard Hugging Face cache (`~/.cache/huggingface/` by default, or `$HF_HOME` if set); this only happens once and requires network access.
 
 Use `semble clear` to remove cached data: `semble clear index` (saved indexes), `semble clear savings` (usage stats), `semble clear orphans` (indexes for repos no longer present on disk), or `semble clear all` (everything).
--- a/src/semble/cli.py
+++ b/src/semble/cli.py
@@ -2,6 +2,8 @@ import argparse
 import asyncio
 import io
 import json
+import logging
+import os
 import re
 import sys
 import warnings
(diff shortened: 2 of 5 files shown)


stephantul:
Please read the [contribution guidelines](github.com/MinishLab/semble/blob/main/CONTRIBUTING.md).

Issue #257, opened 2026-08-29
Gold annotations point at files that don't contain the queried behavior


Hartistic:
I've been running the benchmark against the pinned repos and went through some of the misses by hand. Four gold labels don't hold up once you open the files they point at:

**model2vec — "how vocabulary is pruned during distillation"**
The primary gold is `model2vec/distill/utils.py`, but that file contains a single function, `select_optimal_device()` — it picks cuda/mps/cpu and has nothing to do with vocabulary or pruning. The actual pruning happens via `prune_added_tokens()`, called from `model2vec/tokenizer/tokenizer.py:71` and `model2vec/distill/distillation.py:81`. The latter is already the secondary gold here, so the fix might just be dropping `distill/utils.py` (or swapping in `tokenizer/tokenizer.py`).

**model2vec — "tokenizer construction and vocabulary building"**
Same file again as secondary gold. Still just device selection, no tokenizer or vocabulary content.

**pydantic — "custom field and model validators"**
The secondary gold `pydantic/class_validators.py` is a 5-line V1 migration shim (`__getattr__ = getattr_migration(__name__)`). If the intent was to credit the deprecated V1-style validators, that file is `pydantic/deprecated/class_validators.py`.

**aiohttp — "WebSocket client implementation"**
The secondary gold `aiohttp/_websocket/reader.py` is a 31-line import switch that picks the C or Python reader at import time. The implementation it re-exports lives in `aiohttp/_websocket/reader_py.py`.

All four checked at the SHAs pinned in repos.json and still present in the annotations on main. Happy to put up a PR with whichever corrections you'd prefer — dropping the entries vs. redirecting them changes scores a little either way, so I didn't want to guess at intent.

Hartistic:
Went through the two TypeScript repos the same way. In zod, 8 of the 20 queries point at `packages/zod/src/v4/core/api.ts`, so I checked how much of that file's code actually runs. Of its 114 exported functions, 49 have zero call sites anywhere in the repo (grepped `core.<fn>(` across all of `packages/`, tests included): `classic/schemas.ts` and `mini/schemas.ts` each carry their own independent, identically-behaved builders that construct the schema classes directly instead of calling into `api.ts`. Six of the eight queries land squarely in that unused set; the other two (the general "core public API functions for building schemas", and ".refine/.superRefine") hold up fine and aren't included below.

**zod — "union and discriminated union schema types"**
[...]

PR #259, opened 2026-09-02
Correct eleven gold annotations pointing at dead, shim, or deprecated files

Hartistic:
Follow-up to #257 — the corrections offered there, folded into one PR.

The benchmark scores a tool by whether it finds one specific "right answer" file for each question. For these eleven questions, the file the answer key names isn't where the behavior in the question actually lives: some of the named files hold copies of code nothing in the project calls anymore, some are small forwarding files that only point at the real implementation, and one is an entry point the project itself now warns against using. A tool that returns the genuinely useful file gets marked wrong on these; one that returns the leftover gets full credit. Each replacement below was checked by hand at the revisions pinned in `repos.json` (call-site checks and file reads — the per-line evidence is in the issue thread).

| repo | query | was | now |
|---|---|---|---|
| model2vec | how vocabulary is pruned during distillation | `distill/utils.py` (device selection only) | `tokenizer/tokenizer.py`, where `prune_added_tokens()` lives (secondary `distill/distillation.py` unchanged) |
| model2vec | tokenizer construction and vocabulary building | secondary `distill/utils.py` | secondary dropped |
| pydantic | custom field and model validators | secondary `class_validators.py` (5-line migration shim) | secondary `deprecated/class_validators.py`, the actual V1-style validators |
| aiohttp | WebSocket client implementation | secondary `_websocket/reader.py` (31-line C/Python import switch) | secondary `_websocket/reader_py.py`, the implementation it re-exports |
| vitest | test reporter interface for listening to test events and results | `public/reporters.ts` (re-export barrel, deprecated at runtime since Vitest 4.1) | `node/types/reporter.ts` (the `Reporter` interface itself), secondary `node/reporters/index.ts` |

The six zod queries (union/discriminatedUnion, optional/nullable, transform/pipe, record/map, ZodDefault/ZodCatch, enum/nativeEnum/literal) all share one pattern: `v4/core/api.ts` exports factory helpers that `classic/` and `mini/` reimplement rather than call — 49 of its 114 exported functions have no call site anywhere in the repo. Each of the six now points at `v4/core/schemas.ts` (the runtime classes that do the work) as primary, with `v4/classic/schemas.ts` (the builders users actually reach through `z.*`) as secondary.

Reran semble itself on the five affected repos before and after, same machine, scored the same way as the head-to-head in the issue:

| repo | n | before | after | Δ |
|---|---:|---:|---:|---:|
| aiohttp | 21 | 0.773 | 0.762 | −0.011 |
| model2vec | 20 | 0.716 | 0.738 | +0.022 |
| pydantic | 20 | 0.690 | 0.690 | 0.000 |
| vitest | 20 | 0.654 | 0.641 | −0.013 |
| zod | 20 | 0.558 | 0.639 | +0.081 |
| **overall** | 101 | **0.679** | **0.695** | **+0.016** |

Net positive for semble itself, mostly from zod — pointing the golds at files that actually contain the queried behavior makes them more findable for everyone. I didn't rerun the other models in the published table (as you said, they're the expensive part), so their columns will drift a little on these repos whenever they are.

### DIFF
--- a/benchmarks/annotations/aiohttp.json
+++ b/benchmarks/annotations/aiohttp.json
@@ -21,7 +21,7 @@
       "aiohttp/client_ws.py"
     ],
     "secondary": [
-      "aiohttp/_websocket/reader.py"
+      "aiohttp/_websocket/reader_py.py"
     ],
     "category": "semantic"
   },
--- a/benchmarks/annotations/model2vec.json
+++ b/benchmarks/annotations/model2vec.json
@@ -22,9 +22,7 @@
     "relevant": [
       "model2vec/tokenizer/tokenizer.py"
     ],
-    "secondary": [
-      "model2vec/distill/utils.py"
-    ],
+    "secondary": [],
     "category": "semantic"
   },
   {
@@ -122,7 +120,7 @@
   {
     "query": "how vocabulary is pruned during distillation",
     "relevant": [
-      "model2vec/distill/utils.py"
+      "model2vec/tokenizer/tokenizer.py"
     ],
     "secondary": [
       "model2vec/distill/distillation.py"
--- a/benchmarks/annotations/pydantic.json
+++ b/benchmarks/annotations/pydantic.json
@@ -31,7 +31,7 @@
       "pydantic/functional_validators.py"
     ],
     "secondary": [
-      "pydantic/class_validators.py"
+      "pydantic/deprecated/class_validators.py"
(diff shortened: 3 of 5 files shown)


Pringled (approved):
Hey @Hartistic, thanks for making these fixes, appreciate it! We'll rerun our benchmarks soon with this.

PR #263, opened 2026-09-08
feat: Add multi repo search support

Pringled:
This PR adds support for multi repo search so you can do`semble search "invoice endpoint" ./service-a ../service-b`. This will create (and cache) an index on both paths, and merge them at runtime. For the entire Minish org, this merge takes ~5ms. 

All the agent instructions are also updated so the agent is aware of this new functionality.


src/semble/index/index.py
```diff
@@ -173,6 +177,55 @@ def from_path(
...
+        """Merge indexes built from several repos into one that searches them together.
+
+        :param indexes: (source, index) pairs, where source is the local path or git URL the index was built from.
+        :return: A new in-memory index over all chunks.
+        :raises ValueError: If the indexes were built with different models, or the same repo is passed twice.
+        """
+        if len({index._model_path for _, index in indexes}) != 1:
+            raise ValueError("Indexes to merge must be built with the same model.")
+        # Sort by resolved source so the same set of repos always gets the same labels, whatever the argument order.
+        resolved = sorted(
```
stephantul: Pretty difficult oneliner. The comment explains it, but still, a lot happening. I think you can get away by sorting first, and then doing the list comprehension after (or the other way around)

src/semble/index/index.py
```diff
@@ -173,6 +177,55 @@ def from_path(
...
+            ((s if is_git_url(s) else str(Path(s).expanduser().resolve()), index) for s, index in indexes),
+            key=lambda pair: pair[0],
+        )
+        sources = [source for source, _ in resolved]
+        if len(set(sources)) != len(sources):
+            raise ValueError(f"The same repo was passed more than once: {sources}")
```
stephantul: It is possible to print the actual duplicate here. Maybe a bit too much code for such an edge case though

src/semble/index/index.py
```diff
@@ -173,6 +177,55 @@ def from_path(
...
+        labels: list[str] = []
+        for source in sources:
+            name = label = Path(source).name.removesuffix(".git")
+            suffix = 2
```
stephantul, line 202: I understand why this starts at `2`, but it also looks odd. 

So for repo `a`, which is present twice, we'd get

`a`
`a-2`

This identifier is just internal right? If so, nothing to worry about.

stephantul (approved):
Some small things. I like it

src/semble/index/index.py
Pringled: They are not completely internal, e.g. you do see it in the results so you would see `api-2/src/foo.py` for example. But it's a bit of an edgecase anyway and I don't think there's a much nicer way to fix this besides not indexing (which is worse IMO), wdyt?

src/semble/index/index.py
Pringled: Agreed, will split

src/semble/index/index.py
Pringled: Yeah I like that, done

PR #266, opened 2026-09-11
feat: Fix chunk source line counting, add semble-only benchmark option

Pringled:
`chunk_source` counted newlines in `source[:offset]` for every chunk, re-scanning the file prefix each time, so chunking was quadratic in file size. It now records newline offsets once and uses `bisect_left`, which gives the same line numbers but much faster. This helps mostly on large files which are ~30% faster to index. On average our indexing speed goes up by ~9%.

Also added a `--semble-only` option to the benchmarks for easier reruns when changes just affect semble, and reran the benchmarks and updated our numbers.


stephantul (approved):
WHO WROTE THIS!? (me)

PR #272, opened 2026-09-18
chore: Remove dead code and deduplicate internals

Pringled:
This PR removes a bunch of dead code that was left over from older refactors or features we changed. There's also some deduplication, and `weighting.py` is now just folded into `boosting.py` because it was only a couple of lines and only used there. Scores and speed on the benchmark are the exact same and all tests still pass.


src/semble/installer/agents.py
```diff
@@ -21,11 +21,7 @@ def _exists_or_denied(path: Path) -> bool:
     """Distinguish between existence and permission issues."""
     try:
         path.stat()
-    except FileNotFoundError:
```
stephantul: Is this resolved?

stephantul (approved):
One question about a conspicuous change but otherwise fine.

